PR #48, opened 2026-04-20
fix: split interleaved tiny-text blocks at the same Y

abimaelmartell:
## Summary
- Fixes a layout bug where slide-deck PDFs that stack multiple text blocks into a tiny (<3pt) copy-paste layer at the same Y would extract as character-interleaved gibberish (e.g. `r1e2s1pe8ct3i5ve0o6w1n1e1r4s` for `respective owners`).
- Root cause: the Y-bucket grouper fused every block at that Y into one line, then the X-sort interleaved their glyphs.
- Fix: inside `merge_text_items`, open a new group whenever the incoming item's X is a sharp backtrack from the current bucket's right edge — that's the signature of a new `BT ... Tm` text block restarting at the left. The rule is gated to fonts below 3pt on both sides so normal-size tables and body text keep their existing behaviour.

## Test plan
- [x] `cargo fmt`, `cargo clippy -- -D warnings`, `cargo test` (374 lib + 104 integration + 2 doc).
- [x] New unit tests: `merge_items_splits_tiny_interleaved_text_blocks_at_same_y`, `merge_items_does_not_split_normal_font_table_rows`, plus a kerning-overlap tolerance test so ligatures and subscripts aren't split by accident.
- [x] Verified on the reporting PDF (TPL 8-K investor presentation): the disclaimer paragraph now reads cleanly, matching `pdftotext` output.
- [x] Regression sweep on the pdf-evals snapshot suite (based on the encoding-fix snapshots branch): 39 snapshots changed, 143 unchanged. Spot-check highlights:
  - TPL 8-K disclaimer: garbled → clean prose.
  - ERP-2025: `CO . The economics of2` → `CO2. The economics of` (subscripts merged correctly).
  - HTM_02-01: footnote markers correctly glued to cell values (`80` → `80(3)`).
  - References list (275): broken multi-line citations reflowed to single lines, italic journal titles restored.
  - Revista63: footnote marker glued to text (`procedurii . 8` → `procedurii8.`).
  - A few dense schematic PDFs (ptn3356, MCF5235RM) reorder some component labels — they were dense glyph-by-glyph layouts already, neither clearly better nor worse.
  - Minor: a couple of narrow-column headers reshape the hyphen position (`Struc- tures` → `Struc tures-`).

### DIFF
--- a/src/extractor/mod.rs
+++ b/src/extractor/mod.rs
@@ -315,20 +315,78 @@ pub(crate) fn merge_text_items(items: Vec<TextItem>) -> Vec<TextItem> {
         return items;
     }
 
-    // Group items by (page, Y position) with 5pt tolerance
+    // Group items by (page, Y position) with 5pt tolerance, walking in
+    // stream order.  A new group also starts when the incoming item's X
+    // would be a significant backtrack — signature of a new PDF text
+    // block (`BT ... Tm`) starting at the left of the same Y band.
+    // Without this split, slide-deck PDFs that stack several copy-paste
+    // layers at the same tiny-text Y (e.g. a color-palette row and a
+    // disclaimer body) later sort-by-X into character-interleaved
+    // gibberish.
     let y_tolerance = 5.0;
-    let mut line_groups: Vec<(u32, f32, Vec<&TextItem>)> = Vec::new();
+    struct LineGroup<'a> {
+        page: u32,
+        y: f32,
+        end_x: f32,
+        font_size: f32,
+        items: Vec<&'a TextItem>,
+    }
+    let mut line_groups: Vec<LineGroup> = Vec::new();
 
     for item in &items {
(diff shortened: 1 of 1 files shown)


abimaelmartell:
Closing — the tiny-font gate isn't selective enough in practice. PDFs generated by Microsoft Word / Google Docs use raw Tf sizes of 1pt and scale up via CTM, so both `font_size` and `height` (rendered size) come in below the 3pt gate for normal body text, which then gets mis-split across multi-column layouts. The 2002 IRS form (and ~30 other common PDFs in the eval corpus) shows clear regressions.

A proper fix needs to thread a text-block index (incremented at each `BT`) through `TextItem` so lines can be grouped by (Y, block) rather than inferring block boundaries from X backtracks. That's a bigger change and I'd rather take it on as its own thing than land this half-heuristic.

PR #70, opened 2026-04-28
tables: add detectVectorGridInRegion napi export for region-scoped vector grid detection

abimaelmartell:
## Summary

Adds `detectVectorGridInRegion(pdfBytes, pageIdx, regionPdfPtBbox, renderDpi)` — returns TSR-compatible `{ structureTokens, cellBboxes }` for vector ruled-line / rectangle grids inside a page region, or `null` when no valid grid is detected.

Reuses the existing `extract_page_text_items`, rect detector, and line detector. **No existing API changes behavior.**

## NAPI shape

```ts
detectVectorGridInRegion(
  pdfBytes: Buffer,
  pageIdx: number,
  regionPdfPtBbox: [number, number, number, number],
  renderDpi: number,
): { structureTokens: string[]; cellBboxes: number[][] } | null
```

Example:

```json
{
  "structureTokens": ["<table>", "<tr>", "<td></td>", "<td></td>", "</tr>", "</table>"],
  "cellBboxes": [[0, 0, 80, 30], [80, 0, 160, 30]]
}
```

The structure shape is intentionally compatible with `extractTablesWithStructure*`, so callers can hand the result to the existing TSR pipeline and let the native PDF text-fill path populate cell contents.

## Behavior notes

- Rotated pages (`Rotated90Ccw`) return `null` for v1 with a TODO.
- Detections that would produce out-of-crop cell coordinates return `null` rather than emit invalid bboxes.
- Edge derivation rebuilds boundaries from filtered vector geometry; a TODO is in place to refactor the underlying detectors to expose normalized edges directly.

## Tests

- Synthetic vector-grid PDF (line-drawn): detector emits `{structureTokens, cellBboxes}` and the existing TSR pipeline produces valid markdown when fed the result.
- Synthetic non-table PDF: detector returns `null`.
- Multi-table page with region scoping: only the requested table is detected.
- napi-side smoke test verifying the JS shape.

## Verification

- `cargo fmt --check`
- `cargo test`
- `cargo build --release`
- `cargo clippy --manifest-path napi/Cargo.toml -- -D warnings`
- `cargo test --manifest-path napi/Cargo.toml`
- `npm --prefix napi run build:debug`
- `cd napi && node test.mjs`

This does not publish to npm; the corresponding consumer change and version bump will land separately.

### DIFF
--- a/napi/src/lib.rs
+++ b/napi/src/lib.rs
@@ -99,6 +99,13 @@ pub struct PageRegionTexts {
     pub regions: Vec<RegionText>,
 }
 
+/// Vector-grid detection result compatible with `extractTablesWithStructure*`.
+#[napi(object)]
+pub struct VectorGridDetectionJs {
+    pub structure_tokens: Vec<String>,
+    pub cell_bboxes: Vec<Vec<f64>>,
+}
+
 // ---------------------------------------------------------------------------
 // Helpers
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


abimaelmartell (commented):
(review withdrawn)

abimaelmartell:
Addressed review feedback:

- Added the detector-edge TODO near the export edge reconstruction.
- Changed rotated-page handling to return `null` for v1 with a TODO, instead of emitting unverified transformed bboxes.
- Added crop bbox plausibility validation. The curated smoke initially exposed negative crop bboxes for tight reference table boxes; those now return `null` instead of bad coordinates.
- Reran the curated smoke and pasted both exact-reference-box and full-page results into the PR description.
- Reran verification: `cargo clippy -- -D warnings`, `cargo test`, `cargo build --release`, napi clippy/test, debug napi build, and `node test.mjs`.

abimaelmartell (commented):
(review withdrawn)

abimaelmartell (commented):
Third pass — looks good.

`vector_grid_tests::test_crop_px_bbox_is_plausible_bounds` covers all the requested branches (in-bounds, out-of-bounds on each side, slack boundary, DPI fallback for 0 and negative). Placing the test in-module via `use super::...` rather than in `tests/integration_tests.rs` was the right call — keeps the function private, follows the least-invasive option. CI green across all five jobs. No internal-naming or local-path references remain anywhere in the diff.

LGTM, ready to merge.

PR #71, opened 2026-04-29
tables: expand multi-row cells in-place when fallback heuristic is empty

abimaelmartell:
## Summary
- Adds an in-place recovery path for `multi_row_in_cell` TSR failures before falling back to heuristic extraction.
- Splits overstuffed TSR rows into native PDF text bands and renders those expanded cells as markdown with `multi_row_in_cell_expanded`.
- Covers the behavior with a private unit test plus auto-wrapper integration tests, including an under-counted synthetic vector-grid table.

## Implementation Notes
`multi_row_in_cell` fires in `detect_tsr_quality_issue` at `src/lib.rs:2191`; at that point the code has the `TsrTableInput`, the filled `StructuredCell`s, each cell bbox, page height/rotation, and freshly extracted native `TextItem`s with PDF y-positions. Before this change that state was only used to return the string reason, then the auto path computed a heuristic replacement at `src/lib.rs:2282` by calling `extract_tables_in_regions_mem` over the crop bbox. I chose to reuse the detection pass to cluster native text lines inside each rowspan=1 cell, derive row-level visual bands supported by at least two cells, and clone the TSR row into one row per band while preserving the existing column structure. The main alternative was to keep relying on the heuristic detector, but it often returns empty because its region-level detector has stricter minimum-item/alignment/content validations and can also be handed a crop where no native text matches even though the TSR cell bboxes did. I also avoided adding a parallel detector or changing TableFormer/vector-grid output; this stays in cell-fill/structure recovery and only expands rows where the column structure appears usable.

```text
TSR row:      | Oak Street Boardwalk | 100 200 |
Native bands:       y=Oak/100           y=Boardwalk/200
Expanded:     | Oak Street           | 100 |
              | Boardwalk            | 200 |
```

Columns are treated as already recovered by TSR: the production data showed good token overlap on surviving cells, so this fix does not attempt column recovery. New rows keep a uniform column count by cloning the original row's cells and assigning text per y-band; rows with rowspan/colspan participation are left for the existing heuristic fallback rather than being expanded in v1.

## Fallback Reason Taxonomy
- `None`: clean TSR output, no recovery needed.
- `multi_row_in_cell_expanded`: multi-row-in-cell signal fired and in-place row expansion produced non-empty markdown; heuristic is not consulted.
- `multi_row_in_cell`: expansion could not safely apply, heuristic fallback produced non-empty markdown.
- `multi_row_in_cell_heuristic_empty`: expansion could not safely apply, heuristic returned empty, raw TSR markdown was preserved.
- `multi_row_in_cell_heuristic_error`: expansion could not safely apply and heuristic extraction errored, raw TSR markdown was preserved.
- `phantom_empty_row` and its `_heuristic_empty` / `_heuristic_error` variants continue to use the existing heuristic path.

## Verification
- `cargo fmt --check`
- `cargo test`
- `cargo build --release`

`cargo test` still emits pre-existing test-build warnings in `src/detector.rs` for `font_changes` and `page_has_only_type3_fonts`; clippy with `-D warnings` passes.

Consumer rollout in `fire-pdf` follows separately with a coordinated `@firecrawl/pdf-inspector` version bump. No npm publish or package version bump is included here.

### DIFF
--- a/napi/src/lib.rs
+++ b/napi/src/lib.rs
@@ -482,9 +482,10 @@ pub fn extract_tables_with_structure_cells(
 /// `fallbackReason` is `null` when the TSR-hybrid path produced the
 /// markdown directly. When stage 1's quality check fires (the cells
 /// look like a SLANet detection pathology — phantom rows or multi-row
-/// content in a single cell), the heuristic table extractor is run on
-/// the same region instead, and `fallbackReason` carries the diagnostic
-/// label (`"phantom_empty_row"`, `"multi_row_in_cell"`).
+/// content in a single cell), the auto path may expand the TSR cells
+/// in-place or run the heuristic table extractor on the same region.
+/// `fallbackReason` carries the diagnostic label (for example
+/// `"multi_row_in_cell_expanded"` or `"phantom_empty_row"`).
 #[napi(object)]
 pub struct TableExtractionResultJs {
     pub markdown: String,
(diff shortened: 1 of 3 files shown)


abimaelmartell (commented):
Pass 2 — addressed the band-cap comment and CI is fully green across all five jobs. The 2-line note in `build_tsr_row_expansion` is the right shape: explains the v1 scope cut so a future reader knows the cap is deliberate, not arbitrary. No other changes needed.

LGTM, ready to merge.

PR #78, opened 2026-05-06
tables/detect_rects: don't accept relaxed grid on wireless prose

abimaelmartell:
## Summary
- Reject two-column cell-rect fallback grids when prose-like content is split only by text-position columns rather than rect-derived column geometry.
- Add wireless no-rect regression fixtures covering both full-page rect detection and region-scoped vector-grid detection.
- Bump the NAPI package version to 1.8.6.

## Test plan
- `cargo fmt`
- `cargo test --release`
- `cargo build --release`
- `cd napi && bun install && bun run build`


abimaelmartell:
Validation note, anonymized:

- Wireless test case A: TEDS `0.286` -> `1.000` with the local 1.8.6 build.
- Wireless test case B: TEDS `0.447` -> `0.447` in the local end-to-end route; direct detector probes return no vector grid for its wireless regions, so this PR no longer accepts a rect/vector grid there.
- Rust/NAPI gates passed: `cargo fmt`, `cargo test --release`, `cargo clippy -- -D warnings`, `cargo build --release`, and `cd napi && bun install && bun run build`.

PR #117, opened 2026-07-08
feat(markdown): underline emission, Unicode scripts, style-preserving merges

abimaelmartell:
## Summary

Round 2b of fire-pdf ENG-5015 — the measured residuals after 1.9.9 all lived in the direct-extraction markdown path:

- **`<u>` emission** in `text_with_formatting` (new `detect_underline` option, default on), consuming the geometric `is_underline` flag from #116. Underline runs are kept free of nested `**`/`*` (consumers match tag content literally). Heading lines still emit plain text for bold/italic but preserve `<u>` — the tag carries meaning the `#` prefix doesn't.
- **Unicode scripts at merge time**: `merge_subscript_items` absorbed "H"+"2"→"H2", erasing the script. It now maps digits to Unicode forms with direction from the baseline offset: `H₂`, `word²`, `m³`, `CCl₂F₂`. NFKC/NFKD folds them back to plain digits, so downstream text matching is unaffected.
- **Style-preserving merges**: `merge_text_items` merged adjacent items regardless of bold/italic, absorbing styled runs into plain neighbors (or vice versa) and erasing the flags. It now breaks merge runs at style boundaries. On eval docs this recovers 20–82 italic runs per document that previously emitted as plain text.

## Validation

- New/updated unit tests (script direction, style-boundary merge); full `cargo test` green (542 lib + 138 integration)
- 4 snapshots regenerated — every diff is a feature, reviewed by hand: `CCl₂F₂`/`m³` units, underlined legal section headings in td9264, finer-grained bold runs in p1244
- pdf-evals regression suite: 202/202 real PDFs pass (the one "failure" is a fixture that is literally an HTML error page saved as .pdf, correctly rejected)
- Real eval docs via napi build: underline-bearing docs now emit 4–8 `<u>` spans each on the direct path (was 0); italic-run docs emit 20–82 runs (was 0)

## Known limitations (documented, follow-ups not in this PR)

- Docs whose body renders inside detected pipe-tables (e.g. meeting minutes) lose styling in cell text — table formatting path, separate change
- Synthetic shear-matrix italics (no font flag) still undetected
- napi 1.9.9 → 1.9.10, publishes on merge

---
## Summary by cubic
Adds underline emission, Unicode sub/superscripts, and style-preserving merges to keep formatting in direct Markdown extraction. Cleans up spacing and rejects false-positive row-stripe tables; improves legal headings, scientific units, and lists; addresses ENG-5015 (round 2b).

- **New Features**
  - Emit `<u>` runs via `detect_underline` (default on) and break merges at style boundaries so bold/italic/underline spans stay tight; headers keep plain text for bold/italic but preserve underline.
  - Preserve scripts when merging: digits map to Unicode sub/superscripts based on baseline (e.g., H₂, word², m³); NFKC/NFKD keeps text matching stable.
  - Keep lists valid when underlined by moving bullets outside `<u>` wrappers.
  - Postprocess cleanup: remove a stray space before sentence punctuation and trim spaces next to parentheses/brackets in table cells (e.g., "( twice )" → "(twice)").
  - Table detection: reject sparse two‑column row‑stripe outlines to avoid misclassifying prose as tables.

- **Migration**
  - To disable underline output, set `detect_underline: false`.

<sup>Written for commit e16808c0c4013f4a5c59a2e9a8df47f7e09c7461. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/pdf-inspector/pull/117?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/napi/package.json
+++ b/napi/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@firecrawl/pdf-inspector",
-  "version": "1.9.9",
+  "version": "1.9.10",
   "description": "Fast PDF classification and text extraction. Detect text-based vs scanned PDFs, extract text by region with quality checks. Native Rust performance via napi-rs.",
   "main": "index.js",
   "types": "index.d.ts",
--- a/src/extractor/mod.rs
+++ b/src/extractor/mod.rs
@@ -568,6 +568,13 @@ pub(crate) fn merge_text_items(items: Vec<TextItem>) -> Vec<TextItem> {
                 if (next.font_size - first.font_size).abs() > first.font_size * 0.20 {
                     break;
                 }
+                // Never merge across bold/italic boundaries: the merged item
+                // carries `first`'s flags, so absorbing a styled run into a
+                // plain neighbor (or vice versa) silently erases the styling
+                // that markdown emission and downstream inline-styling need.
+                if next.is_bold != first.is_bold || next.is_italic != first.is_italic {
+                    break;
+                }
                 let gap = next.x - end_x;
                 let x_gap_max = if *preserve_stream_order && is_standalone_bullet_text(&text) {
                     first.font_size * 1.2
@@ -710,7 +717,17 @@ pub(crate) fn merge_subscript_items(items: Vec<TextItem>) -> Vec<TextItem> {
                         let gap = item.x - parent_right;
                         // Subscripts must be tightly adjacent (within ~1pt)
                         if gap < parent.font_size * 0.2 && gap > -parent.font_size * 0.3 {
-                            parent.text.push_str(&item.text);
+                            // Preserve the script when absorbing it: map the
+                            // digits to Unicode sub/superscript forms so the
+                            // raised/lowered rendering survives in extracted
+                            // text ("H"+"2" → "H₂", "word"+"2" → "word²").
+                            // NFKC/NFKD normalization folds these back to
+                            // plain digits, so text matching downstream is
+                            // unaffected. Direction from the baseline offset
+                            // (y-up here): raised → superscript (footnote
+                            // refs), lowered/level → subscript (chemistry).
+                            let raised = item.y > parent.y + parent.font_size * 0.1;
+                            parent.text.push_str(&map_script_digits(&item.text, raised));
                             parent.width = (item.x + item.width) - parent.x;
                             continue;
                         }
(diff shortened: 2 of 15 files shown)


abimaelmartell:
Both addressed:

- **Underline stretching across merges (P2)**: fixed as suggested — merge runs now break on `is_underline` changes too (any style-flag change), and the OR-accumulator is gone (redundant once merges can't mix flags). The pre-existing `merge_items_preserves_underline_from_later_fragment` test encoded the OR semantics; updated to assert the fragments stay separate (line-level assembly still joins them without a space, so rendered text is unchanged — `pre<u>fix</u>`). The td9264 snapshot got tighter, more accurate spans out of this: trailing periods now sit correctly outside `<u>`.
- **Underlined bullet lists (P2)**: `format_list_item` now moves bullet markers outside `<u>` wrappers (same treatment as the existing `**`/`*` handling), with a unit test — `<u>● Item</u>` → `- <u>Item</u>`.

Re-validated: full cargo suite green (544 lib + 138 integration), and the pdf-evals regression corpus re-ran with the updated binary — the only content diffs across 203 PDFs are the intended styling changes (`<u>` on genuinely underlined headings, italic runs, Unicode scripts).

abimaelmartell:
Fixed. The "word ." artifacts came from trailing-period fragments stranded by the style-boundary item splits and joined with spaces by line/cell assembly. Rather than patching each join site (the artifact surfaced through several: paragraph joins, table cell joins, footnote joins), a single postprocess pass now strips a space before `.`/`,`/`;` when the mark ends its token (followed by whitespace, a `|` cell boundary, or end of text). Dot leaders / ellipses and mid-token periods are explicitly protected, with unit tests for all four cases.

td9264 is now artifact-free, and the fix also cleaned two pre-existing "armoring ," artifacts in the 2013-app2 snapshot. Re-ran the pdf-evals corpus: zero markdown diffs across all 203 PDFs vs committed baselines (the suite's similarity-score "FAIL"s are metadata timing noise — byte-identical markdown, verified by direct diff).

abimaelmartell:
@cubic-dev-ai

abimaelmartell:
@cubic-dev-ai

abimaelmartell:
@cubic-dev-ai

PR #123, opened 2026-07-10
ci: add PyPI trusted publishing (abi3 wheels, v0.2.1)

abimaelmartell:
## Summary

Sets up automated PyPI publishing so `pdf-inspector` releases stop depending on manual uploads from a personal machine. Same trusted-publishing (OIDC) pattern as our npm and crates.io workflows — **no API tokens anywhere**.

- **`publish-pypi.yml`**: triggers when the `[project]` version in `pyproject.toml` changes on main — same flow as npm publishing keying off `napi/package.json`. Builds wheels for linux x86_64/aarch64, macOS x86_64/arm64, windows x64 + sdist via maturin, publishes with `pypa/gh-action-pypi-publish` using OIDC. `workflow_dispatch` (main-only) is a manual fallback that also repairs partially-uploaded releases via `skip-existing`.
- **Independent Python version**: `pyproject.toml` now carries its own static `version` (set to 0.2.1 — PyPI is already at 0.2.0, published outside this repo, so anything ≤0.2.0 would be ignored by pip). Bump it manually to release; the crate version in Cargo.toml is untouched by this PR.
- **abi3 wheels** (`abi3-py38`): one wheel per platform now covers all CPython ≥3.8. The previous manual uploads were cp312-only, so users on 3.10/3.11/3.13 were building from source.

## Verification

- `cargo fmt` / `cargo clippy --features python -- -D warnings` / `cargo test` all pass
- Built the wheel locally: `pdf_inspector-0.2.1-cp38-abi3` (static pyproject version wins over Cargo's 0.1.4), type stubs (`pdf_inspector.pyi`) included by maturin
- Installed in a clean venv: `tests/test_python.py` — 63 passed
- Extraction no-op confirmed: binaries built from main with and without this change produce byte-identical output on all 41 fixture + eval PDFs


abimaelmartell:
Both review findings addressed in 89b20d5:

1. **Branch guard**: `check-version` (which gates all downstream jobs) now requires `github.ref == 'refs/heads/main'`, so a workflow_dispatch from a feature branch can no longer publish unmerged code.
2. **Partial-release repair**: manual dispatch now always rebuilds instead of trusting the PyPI version check (which returns 200 even for incomplete releases), and the publish step uses `skip-existing: true` so a re-run uploads only the missing files.

PR #125, opened 2026-07-10
feat(extractor): descriptor/embedded-font style flags + geometric strikeout detection

abimaelmartell:
Two style-recall gaps, both invisible to the name-based bold/italic heuristics:

## 1. Descriptor + embedded-font style flags

Subset fonts with opaque `BaseFont` names (`Tc1`, `AAAAAB+Amplitude`) defeat `is_italic_font`/`is_bold_font`. New `descriptor_style_flags`:
- **FontDescriptor**: `ItalicAngle` beyond ±4° (skips token optical slants), Flags bit 7 (Italic), bit 19 (ForceBold). Type0 fonts resolve through `DescendantFonts[0]`.
- **Embedded-font fallback when the descriptor claims upright** (descriptors lie): `ttf_parser::Face` `is_italic()` / `italic_angle()` for sfnt fonts; for bare-CFF `FontFile3` (which ttf-parser can't open — "unknown magic") a minimal CFF Name INDEX parser recovers the real PostScript name and feeds it to the existing name heuristics. Observed in the wild: descriptor rewritten to `ItalicAngle 0` while embedding `Amplitude-LightItalic`.
- ORed into `is_bold`/`is_italic` at item creation (content streams + form XObjects) — flags can only be gained, never lost.

## 2. `is_strikeout` on TextItem

Same geometric pass as underline detection (stroked lines / thin filled rects, all the table-ruling suppression), different vertical window: a rule crossing the glyphs at **12–55% of the em above the baseline** instead of sitting at it. Exposed through napi + python bindings and `pdf2md --items-json`; table-cell suppression in `suppress_table_underlines` covers both flags.

## Verification (public ParseBench corpus)

- Previously-missed italic council titles and bold CJK itinerary headings now flagged — **render-checked** against the actual page images.
- Corpus sweep over 508 text docs: 24 docs gain flags, **none lose any**; spot-checked the largest movers visually (bold reviewer-response answers with bold-italic quotes; bold travel-itinerary headings; italic paper affiliations) — all correct.
- 35 strikeout items detected corpus-wide, disjoint from underline on the same docs.
- 10 new unit tests (4 strikeout windows, 6 descriptor/CFF paths incl. Type0 descendant resolution and the tiny-angle guard). Full suite: 575 + 139 pass, clippy `-D warnings` clean.

Follow-up (separate): fire-pdf emits `~~`/`<s>` from the new flag and bumps the package (ENG-5028).

https://claude.ai/code/session_01L3U6BKYCS73DVA83odAfYB

---
## Summary by cubic
Recover bold/italic for subset fonts using descriptor and embedded font data, and add geometric strikeout detection with a new `is_strikeout` flag across outputs. Improves text layout correctness (quote-op advance, text rise, ActualText positioning) and prevents incorrect style merges, without breaking APIs.

- **New Features**
  - Descriptor-based bold/italic: read `FontDescriptor` (`ItalicAngle` ≥ 4°, Flags bit 7 Italic, bit 19 ForceBold), resolve Type0 via `DescendantFonts[0]`, and fall back to embedded font data (`ttf_parser` for sfnt; CFF Name INDEX for bare `FontFile3`). ORed into name heuristics for items from content streams and form XObjects.
  - Strikeout detection: set `is_strikeout` when a rule/thin rect crosses glyphs 12–55% above the baseline; shares the underline pipeline and table-ruling suppression. Exposed via `napi`, Python, and `pdf2md --items-json`.

- **Bug Fixes**
  - Text layout: `'` now carries advance width and advances the text matrix; `Ts` rise shifts rendered positions; ActualText items render at their glyphs’ rise and capture position after the `'` line move.
  - Performance: document-level `FontStyleCache` memoizes embedded font style flags by FontFile2/3 object id to avoid repeated decompression/parsing.
  - Merging and suppression: subscript merge now respects underline/strikeout style boundaries; table suppression clears `is_strikeout` alongside `is_underline`.

<sup>Written for commit f7993077ab1712bd1002edcf3db9e283b462a186. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/pdf-inspector/pull/125?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/napi/src/lib.rs
+++ b/napi/src/lib.rs
@@ -83,6 +83,9 @@ pub struct TextItem {
     /// Underline detected geometrically (drawn rule/thin rect under the
     /// baseline) — PDFs carry no underline font flag.
     pub is_underline: bool,
+    /// Strikeout detected geometrically (rule crossing the glyphs at mid
+    /// x-height).
+    pub is_strikeout: bool,
     pub item_type: ItemType,
     /// URL for link items, `None` for other types.
     pub link_url: Option<String>,
@@ -294,6 +297,7 @@ pub fn extract_text_with_positions(
                     is_bold: item.is_bold,
                     is_italic: item.is_italic,
                     is_underline: item.is_underline,
+                    is_strikeout: item.is_strikeout,
                     item_type,
                     link_url,
                 }
--- a/pdf_inspector.pyi
+++ b/pdf_inspector.pyi
@@ -39,6 +39,7 @@ class TextItem:
     is_bold: bool
     is_italic: bool
     is_underline: bool
+    is_strikeout: bool
     item_type: str
 
 class RegionText:
--- a/src/bin/pdf2md.rs
+++ b/src/bin/pdf2md.rs
@@ -74,7 +74,7 @@ fn format_items_json(items: &[TextItem]) -> String {
                 _ => String::new(),
             };
             format!(
-                r#"{{"text":"{}","page":{},"x":{:.2},"y":{:.2},"width":{:.2},"height":{:.2},"font":"{}","font_size":{:.2},"is_bold":{},"is_italic":{},"is_underline":{},"item_type":"{}","mcid":{}{}}}"#,
+                r#"{{"text":"{}","page":{},"x":{:.2},"y":{:.2},"width":{:.2},"height":{:.2},"font":"{}","font_size":{:.2},"is_bold":{},"is_italic":{},"is_underline":{},"is_strikeout":{},"item_type":"{}","mcid":{}{}}}"#,
                 json_escape(&item.text),
                 item.page,
                 item.x,
@@ -86,6 +86,7 @@ fn format_items_json(items: &[TextItem]) -> String {
                 item.is_bold,
                 item.is_italic,
                 item.is_underline,
+                item.is_strikeout,
                 item_type_label(&item.item_type),
                 mcid,
                 link_url,
@@ -122,6 +123,7 @@ mod tests {
             is_bold: false,
             is_italic: true,
             is_underline: true,
+            is_strikeout: true,
             item_type: ItemType::Text,
             mcid: Some(7),
         }];
--- a/src/extractor/content_stream.rs
+++ b/src/extractor/content_stream.rs
@@ -14,8 +14,8 @@ use lopdf::{Document, Encoding, Object, ObjectId};
 use std::collections::HashMap;
 
 use super::fonts::{
-    build_font_encodings, build_font_widths, compute_string_width_ts, extract_text_from_operand,
-    get_font_file2_obj_num, get_operand_bytes, CMapDecisionCache,
+    build_font_encodings, build_font_widths, compute_string_width_ts, descriptor_style_flags,
+    extract_text_from_operand, get_font_file2_obj_num, get_operand_bytes, CMapDecisionCache,
 };
 use super::underline::UnderlineLine;
 use super::xobjects::{extract_form_xobject_text, get_page_xobjects, XObjectType};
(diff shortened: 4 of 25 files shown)


abimaelmartell:
Addressed the cubic review in bfe6d5c — all four comments validated against the code first:

1. **`'` operator zero width — VALID, pre-existing (also broke underline) — FIXED.** The `'` handler pushed `width: 0.0` and never advanced the text matrix, so `'`-shown text could never match either detection window, and following text overlapped it. Now reuses Tj's `compute_string_width_ts` path and advances the matrix. Tests: `quote_operator_text_carries_advance_width`, `quote_operator_advances_text_matrix`.

2. **Form XObject rules never reach detection — VALID but pre-existing (since #116) — deferred to #126.** The form walker handles no path/paint operators at all and returns only text items; fixing it needs path state, painted-vs-clip discrimination, form-matrix × CTM stroke transforms, and merge plumbing that doesn't exist today. Scoped follow-up: #126.

3. **Repeated embedded-font parsing — VALID — FIXED.** Added a document-scoped `FontStyleCache` keyed by the FontFile2/FontFile3 ObjectId, threaded like the existing `CMapDecisionCache`. Test proves a warm cache answers without re-reading the stream.

4. **`Ts` text rise — VALID, pre-existing (also affected underline) — FIXED.** No `Ts` handler existed; `TextItem.y` ignored rise entirely. Now tracked in the graphics state (saved/restored with q/Q), applied at the three emission sites; advances stay on the unshifted matrix per the rendering-matrix definition. Tests: `text_rise_shifts_item_baseline`, `strikeout_detected_on_risen_text`.

fmt / clippy `-D warnings` / full test suite (580 lib + 139 integration) pass; sanity-ran `pdf2md` on 4 real fixtures.

abimaelmartell:
Valid — fixed in 2908273. The EMC-built ActualText item used the captured matrix without `rise_adjusted`, so tagged runs shown with `Ts` landed on the unshifted baseline while ordinary runs (fixed earlier in this PR) render risen. The rise is now captured together with the first-glyph matrix (and at BDC for the entry-position fallback) — the item renders at the rise of its glyphs, not whatever rise is in effect at EMC. Regression test: `actual_text_item_uses_glyph_rise` (ActualText run at `5 Ts` asserts y=505 while the following untagged run returns to 500).

abimaelmartell:
Valid — fixed in d6d0370. The `'` handler skipped the whole suppressed-extraction block, so a tagged span shown with `'` never captured its glyph matrix/rise and the EMC item fell back to the BDC-entry matrix — on the *previous line* (the `'` line move happens after BDC), unrisen. The capture now happens right after the line move, matching Tj/TJ. Regression test: `actual_text_shown_with_quote_op_uses_moved_risen_baseline` (14 TL line move + 3 Ts → item at y=489, width > 0).

abimaelmartell:
@cubic-dev-ai

abimaelmartell:
Latest review round addressed in f799307:

- **Subscript-merge strikeout loss (content_stream.rs:542) — VALID (pre-existing for underline too) — FIXED.** `merge_subscript_items` absorbed the script digit regardless of underline/strikeout flags — dropping the digit's mark or widening the parent's over it. Differing marks now break the merge, mirroring `merge_text_items`' style-boundary rule. Test: `subscript_digit_with_different_marks_is_not_absorbed` (differing marks stay separate, matching marks still merge).
- **Table-cell strikeout suppression coverage (mod.rs:272) — FIXED.** The suppression test now sets and asserts `is_strikeout` alongside `is_underline`.
- **Form XObject rules (xobjects.rs:309)** — same finding as the previous round; tracked as #126 (needs path/paint state, painted-vs-clip discrimination, and form-matrix × CTM transforms that don't exist in the form walker today — pre-dates this PR, from #116).

fmt / clippy `-D warnings` / full suite (583 lib + 139 integration) pass.

abimaelmartell:
@cubic-dev-ai

PR #131, opened 2026-07-11
fix(markdown): ToC-page suppression, wrapped bold headings, math fragments

abimaelmartell:
## Summary

Three targeted heading-classification improvements, applied to both converter paths:

**Suppress heading promotion on table-of-contents pages.** ToC entries without dot leaders ("1. Overview of OCR Pack", "Part III: Commercial Geographies of Play") are structurally indistinguishable from headings, so whole contents pages came out as stacks of `##`. After emitting a heading whose text is "Contents"/"Table of Contents", heuristic heading promotion is disabled for the remainder of that page (struct-tree headings are unaffected).

**Merge wrapped bold headings.** `merge_heading_lines` only considered font-size-tier and struct-tree headings, so a bold-at-body-size heading that wraps produced two separate headings ("## …of wood pellets and cost" / "## structure in Japan"). A fully-bold line now merges into the previous fully-bold line when it reads as a wrap continuation: starts lowercase, tiny Y gap, and the previous line has no terminal punctuation. Kept deliberately narrow — bold labels and bold sentences start with capitals/markers and are unaffected.

**Reject display-math fragments.** Equation lines ending in an equation number ("S = kB ln W, (2)") and equation lead-ins ("Rearranging Equation (8) gives:") scored as headings via the rarity heuristic — display math has unusual fonts and large gaps, which are exactly the heading signals. The discriminator is an `(N)` equation reference. A bare trailing colon is deliberately **not** a fragment signal: real headings frequently end with colons ("Procedure:", "Steps for Using the Microscope:") — an earlier draft of this rule rejected those and it showed up immediately as regressions, so the tests pin both directions.

The `p1244-1996` snapshot change is the bold-merge working as intended: the stacked IRS form labels "**Subtotals** **from pages**" now read "**Subtotals from pages**".

## Testing

- Unit tests for all three behaviors, including the colon-heading counterexamples and non-merge cases (bold sentences, distinct bold labels)
- `cargo fmt` / `cargo clippy -- -D warnings` / `cargo test` pass (728 tests)
- pdf-evals: 202/203 pass (the 1 failure is the known corrupt fixture, an HTML file with a .pdf extension)

---
## Summary by cubic
Improves heading detection to avoid false headings and split titles. Suppresses ToC-page promotions, safely merges wrapped bold headings, and filters display‑math and running‑header fragments across all heuristics.

- **Bug Fixes**
  - Suppress heading promotion for the rest of a page after a "Contents"/"Table of Contents" heading to prevent ToC entries becoming headings.
  - Merge fully bold wrapped lines only when both lines are fully bold and tier‑less, the second starts lowercase, the Y gap is small, and the previous line ends without punctuation; protects real tiered headings from absorbing bold body text.
  - Exclude display‑math and page‑of‑total fragments from heading heuristics: require math evidence for "(N)" suffixes (operator or comma/colon before the number), keep colon‑ended headings valid, and allow name+year/appendix headings like "Nicaea (325)".

<sup>Written for commit 4b1c7b21b6c0f8f4bd8cfeb91d3b4a5846997833. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/pdf-inspector/pull/131?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/markdown/analysis.rs
+++ b/src/markdown/analysis.rs
@@ -164,9 +164,13 @@ pub(crate) fn is_toc_marker_heading(text: &str) -> bool {
 /// Lines that resemble headings structurally but are display-math fragments:
 /// equations ending in an equation number ("S = kB ln W, (2)") or equation
 /// lead-ins ("Rearranging Equation (8) gives:"). Both carry an "(N)" equation
-/// reference — that's the discriminator. A bare trailing colon is NOT a
-/// fragment signal: real headings frequently end with colons ("Procedure:",
-/// "Steps for Using the Microscope:").
+/// reference — but a trailing "(N)" alone is not enough: real headings end
+/// with parenthesized numbers too ("Nicaea (325)", appendix numbering), so
+/// the suffix form additionally requires math evidence — an "=" in the line
+/// or a comma immediately before the number, both present in every display
+/// equation and absent from name-plus-number headings. A bare trailing colon
+/// is NOT a fragment signal either: real headings frequently end with colons
+/// ("Procedure:", "Steps for Using the Microscope:").
 pub(crate) fn is_heading_fragment(text: &str) -> bool {
     let t = text.trim_end();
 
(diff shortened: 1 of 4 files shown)


abimaelmartell:
All three findings addressed in the latest commit, each checked against the corpora rather than applied blind:

1. **Tier path unguarded** — valid; `is_heading_fragment` is hoisted into the outer gate so both the tier and rarity paths are covered.

2. **Bold heading absorbing bold body** — valid, and the corpus confirmed it concretely: the old merge was gluing a wrapped list-item fragment onto a real `##` heading; with the suggested prev-tier-less condition the output now matches the reference extraction exactly. Added a test pinning the tiered-heading + bold-body non-merge.

3. **Bare `(N)` suffix too aggressive** — valid: the corpora contain real headings ending in parenthesized numbers ("Nicaea (325)", "Chalcedon (451)", Arabic appendix numbering) that the old rule suppressed. The rule now requires math evidence — an operator (`=`, `≤`, `≪`, …) in the line or a `,`/`:` immediately before the number, present in every display equation and absent from name-plus-number headings. Two follow-ons surfaced by the corpus diff are handled: comparison-operator-only equations (`PLL≪PHL: (12)`) and page-of-total running headers ("PM 2 (10)"), which the old blanket rule had been suppressing only by accident and now have an explicit rule.

Full-corpus diff reviewed for every changed document; pdf-evals 202/203 (known corrupt fixture); all tests green.

PR #132, opened 2026-07-11
fix(markdown): keep isolated headings on sparse pages; gate tagged roles

abimaelmartell:
## Summary

**Sparse-page isolation.** `find_isolated_lines` has a density guard that wipes all isolated lines on any page where they exceed 25% of the page's lines — a defense against multi-column layouts where every wrapped line looks isolated. But on a sparse page (a cover, a section divider, a table-of-contents page with a lone "CONTENTS" title and a couple of entries) one isolated line is trivially over 25%, so the guard erased the single heading it was meant to detect. The guard now requires the page to have ≥10 lines before it runs; the ratio only means "multi-column misfire" on a dense page.

**Tagged-role gating.** Letting those isolated lines through surfaced a latent gap: the visual heading heuristic didn't check struct-tree roles, so a line explicitly tagged as a list item, blockquote, code, caption, or ToC entry — or set in a monospace font — could be promoted to a heading if it happened to look isolated. The heuristic is now gated on those roles in both converter paths (explicit heading tags are unaffected).

## Testing

- Unit tests: sparse-page title kept isolated, dense page of isolated lines still wiped, plus the existing tagged-role tests (list/quote/code) that this change is careful not to break
- `cargo fmt` / `cargo clippy -- -D warnings` / `cargo test` pass
- pdf-evals: 202/203 (known corrupt fixture); the `p1244-1996` snapshot update is an improvement — a wrapped "Instructions (continued)" heading now renders as one heading

---
## Summary by cubic
Keep isolated headings on sparse pages and prevent non-heading tagged content—including table cells—from being promoted. Fixes false negatives on covers/ToC pages and false positives for lists, quotes, captions, code, ToC/Index, notes/refs, formulas/forms, and table headers/cells.

- **Bug Fixes**
  - Only run the isolated-line density guard on pages with >=10 lines. Keeps lone titles on sparse pages; still wipes multi-column misfires.
  - Gate visual heading promotion using `StructRole::is_non_heading_content` and monospace font detection. Blocks roles: lists (L/LI/Lbl/LBody), quotes, captions, ToC/TOCI/Index, notes/references/bib entries, code, formulas, forms, and table roles (Table/TR/TH/TD/THead/TBody/TFoot). `Figure` remains allowed. Applied in both converter paths; explicit heading tags are unaffected.
  - Snapshot update: wrapped “Instructions (continued)” now renders as a single heading.

<sup>Written for commit 573bc3cc90428d5530c06c1eda097bad6cbe5a95. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/pdf-inspector/pull/132?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


abimaelmartell:
Valid — extended the gate. Moved it to `StructRole::is_non_heading_content` (testable, single source of truth) and added the missing content roles: **Quote, Index, Note, Reference, BibEntry, Formula, Form**, on top of the existing list/quote/caption/ToC/code roles. `P` and the generic container/flow roles remain excluded so the heuristic still fires there.

One exception from the suggested list: **Figure is deliberately left out.** Adding it caused a real regression — the LA County First Responder Protocol cover page tags its document title ("LOS ANGELES COUNTY" / "LAW ENFORCEMENT FIRST RESPONDER PROTOCOL") inside a Figure alongside the county seal, and the reference extraction renders those as `#`/`##`. Cover and banner pages do this routinely, so a blanket Figure block demotes real titles to bold. The other seven roles are inert across both regression corpora (they never false-fired), so they're pure safety with no downside; Figure was the only one with measurable behavior, and it was negative.

Verified: full corpus diff is exactly one document with Figure included (the regression) and zero with it removed; all tests + clippy green; pdf-evals 202/203.

abimaelmartell:
Valid — added the table roles (**Table, TR, TH, TD, THead, TBody, TFoot**) to `is_non_heading_content`. A `TH` in particular is heading-like enough (short, often bold) to trip the heuristic if table reconstruction falls back and the cell reaches the line loop as plain text.

Verified defensive: the full-corpus diff (400+ PDFs, both suites) is unchanged with these roles added — no tagged-table-fallback case exists in the corpora today, so this is pure hardening for that path with zero behavioral risk. Test extended to cover all seven roles; clippy + tests green.

PR #138, opened 2026-07-11
feat: --password support for encrypted PDFs

abimaelmartell:
## Summary

Adds password support for encrypted PDFs.

- **`PdfOptions.password`** (builder `.password("…")`) threaded through the document loader into both `process_pdf_with_options` and `process_pdf_mem_with_options`.
- **`pdf2md --password PW`** CLI flag.
- **Latent bug fix**: some encrypted PDFs load structurally via `load_mem` with `is_encrypted()` still `true` — the streams stay encrypted, so the extracted "text" is garbage and the file gets misclassified as Scanned. The loader previously only re-decrypted when `load_mem` *errored*. It now also re-decrypts when the loaded document reports `is_encrypted()`, falling back to the empty password (owner-only encryption) when a password is absent or rejected.

Behavior on the encrypted fixture: no password → `Encrypted` error; wrong password → `Encrypted` error; correct password → decrypts and extracts real content.

## Testing

- New `tests/fixtures/encrypted-secret123.pdf` + integration test covering all three cases.
- `cargo fmt` / `cargo clippy -- -D warnings` / `cargo test` pass.
- No behavior change on non-encrypted PDFs: binaries built from `main` with and without this change produce byte-identical output across all 200 regression corpus PDFs.

---
## Summary by cubic
Add password support for encrypted PDFs in both the library and CLI. Also fix a loader bug where some encrypted PDFs loaded with encrypted streams, causing garbled text and misclassification.

- **New Features**
  - `PdfOptions.password` with builder `.password("...")`, passed through to both memory and file loaders.
  - `pdf2md --password PW` flag.
  - Encrypted behavior: no or wrong password returns `Encrypted`; correct password decrypts and extracts real content.

- **Bug Fixes**
  - If `load_mem` succeeds but `doc.is_encrypted()` is true, re-load with the provided password, then fall back to the empty password.
  - Redact password in `PdfOptions` Debug to avoid leaking secrets in logs or panics.
  - No changes for non-encrypted PDFs; outputs are byte-identical across the regression corpus.

<sup>Written for commit 0159910b5f8d268e83f7011c4a72fcaf32e24335. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/pdf-inspector/pull/138?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


abimaelmartell:
Both addressed:

1. **Password leak via Debug** — valid. `PdfOptions` now has a manual `Debug` impl that renders the password as `Some("[REDACTED]")` (all other fields unchanged, `Clone` retained), so logging or a panic that formats the options can't leak the credential. Added `pdf_options_debug_redacts_password` to lock it.
2. **Weak decrypt assertion** — valid. Removed the `|| md.len() > 200` fallback; the test now asserts the fixture's `Procurement` token only, so a garbled-but-long extraction (exactly the encrypted-stream regression this guards) fails the test.

fmt/clippy/tests green.

PR #154, opened 2026-07-15
feat(npm): split platform binaries into optionalDependencies (1.11.0)

abimaelmartell:
## Summary

`@firecrawl/pdf-inspector` ships all three platform binaries in one package — **17.6 MB unpacked**, so every install downloads every platform's binary. This splits it the standard napi-rs way:

- **3 platform packages** (`@firecrawl/pdf-inspector-linux-x64-gnu` / `-darwin-arm64` / `-win32-x64-msvc`), each containing only its `.node` binary with `os`/`cpu`(/`libc`) fields so npm installs exactly one. Generated and published by the workflow — names match the napi-generated loader's existing `require` fallbacks.
- **Main package** drops `*.node` from `files` → **8.5 kB tarball / 44.8 kB unpacked**, and pins the platform packages as `optionalDependencies`. The publish job re-stamps the pins to the exact version so they can never drift on future bumps.
- **Workflow hardening**: `workflow_dispatch` fallback (main-only guard, same as crates/PyPI) and per-package `npm view` already-published checks, so a partial release is retryable idempotently.

## Verification (local, darwin-arm64)

- Built the addon; generated loader falls back to `@firecrawl/pdf-inspector-darwin-arm64`. `npm pack --dry-run`: main tarball = index.js, index.d.ts, bin/, README (no binaries).
- Simulated the workflow's platform-package generation, packed both tarballs, installed them in a clean project: `processPdf` returns identical output to the bundled 1.10.4 (11,118 md chars on the fixture), binary resolves from the platform package, and the `pdf-inspector` CLI bin works.
- `bun install` with a cold cache succeeds while the platform packages are still unpublished (optional deps tolerated), so the build job has no chicken-and-egg problem.

## ⚠️ One-time setup before this can publish

npm **trusted publishing can't create brand-new packages** — the 3 platform packages must exist and have this workflow configured as trusted publisher on npmjs.com before the publish job can succeed. See PR discussion.

---
## Summary by cubic
Split `@firecrawl/pdf-inspector` into platform-specific binary packages listed as `optionalDependencies`, so installs fetch only the needed binary and the main tarball is ~8.5 kB. Hardened the publish workflow with per-package checks, exact version pinning, and a manual retry path.

- **Refactors**
  - Added `@firecrawl/pdf-inspector-{linux-x64-gnu,darwin-arm64,win32-x64-msvc}` packages, each shipping a single `.node` with proper `os`/`cpu`/`libc`.
  - Main package drops `*.node` from `files` and pins the platform packages in `optionalDependencies` to the exact version at publish.
  - Publish flow: guards `workflow_dispatch` to `main`, skips already-published packages, publishes platforms first, then the main package.
  - Docs: README now lists Windows support and the platform package mapping; removed stale `napi.package.name` to align with loader/package names.

- **Migration**
  - Pre-create the three platform packages on npm and enable trusted publishing for this workflow before the first release.

<sup>Written for commit 20d6540fb55f352cb403277e05effeb111834258. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/pdf-inspector/pull/154?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -4,6 +4,9 @@ on:
   push:
     branches: [main]
     paths: ['napi/package.json']
+  # Manual fallback: retry a publish that failed partway (per-package
+  # already-published checks make re-runs idempotent).
+  workflow_dispatch:
 
 permissions:
   contents: read
@@ -12,6 +15,10 @@ permissions:
 jobs:
   check-version:
     name: Check version change
+    # Guard manual dispatches: npm trusted publishing matches
+    # repo+workflow+environment but NOT branch, so without this a
+    # workflow_dispatch from any branch could publish unmerged code.
+    if: github.ref == 'refs/heads/main'
     runs-on: ubuntu-latest
     outputs:
       changed: ${{ steps.check.outputs.changed }}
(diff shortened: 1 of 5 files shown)


abimaelmartell:
Both review findings addressed:

1. **Windows docs**: `napi/README.md` now lists all three platforms (including Windows x64, which has shipped binaries since before this PR) and documents the per-platform packages installed via `optionalDependencies`.
2. **Stale `napi.package.name`**: removed `@firecrawl/pdf-inspector-js` from the napi config — the generated loader and the publish workflow both derive platform package names from the root package name. Verified the generated `index.js` is byte-identical on the fallback `require` names with the config removed.

Issue #118, opened 2026-07-09
Garbled text (systematic per-range character shift) on Type0 fonts with ToUnicode present — extracted silently, never flagged for OCR


abimaelmartell:
## Summary

`text_simple__att10k.pdf` (ParseBench eval corpus) extracts as consistent mojibake: every character comes out shifted by a constant per character class (lowercase −15, uppercase −11, digits/punctuation by other fixed offsets). The document is classified TEXT-BASED, `ToUnicode` is present (`toUni=true`), no decode error is raised, and **neither `is_garbage_text` nor `detect_encoding_issues` flags it** — so `needs_ocr` stays false and the garbage is served silently. Downstream (fire-pdf / ParseBench) this doc fails every content and formatting rule (28 `is_underline` rules alone) despite structurally perfect extraction — even the `<u>` underline spans emit correctly around the garbled text.

## Reproduction

`text_simple__att10k.pdf` from the ParseBench dataset (HuggingFace `llamaindex/ParseBench`, `docs/text/text_simple__att10k.pdf`). Local copy: `~/Code/fire-pdf/evals/parsebench/vendor/ParseBench/data/docs/text/text_simple__att10k.pdf` (sha256 prefix `8b9aecf13c2eeee8`). It is an AT&T 10-K exhibit index — plenty of ground truth text to compare against.

```bash
cargo run --release --bin pdf2md -- text_simple__att10k.pdf | head -5
# 3FFQ&
# -'V 8VceZWZTReVW9VdZXReZdhZeYcVdaVTeeHVcZVd8EcVWVccVUHeT:iYZSZe-(,e2'WZ]V ;VScfRcj2&,*,*$ ...
```

`detect-pdf` says `Type: TEXT-BASED (extractable text)`; `extractPagesMarkdown` returns `pagesNeedingOcr: []`, `ocrReasonsByPage: []`.

## Evidence of the shift pattern

Extracted vs actual (from the rendered page):

| extracted | actual |
|---|---|
| `8VceZWZTReV` | `Certificate` |
| `9VdZXReZ...` | `Designati...` |
| `:iYZSZe` | `Exhibit` |
| `;VScfRcj2&,*,*$` | `February 18, 2020)` (approx.) |

[...]

PR #209, opened 2026-08-02
fix(tounicode): skip subset GID remap for CIDFontType0 (CFF) descendants

### LINKED ISSUE #208: Subset GID remap corrupts CIDFontType0 (CFF) text — Japanese Adobe-Japan1 PDFs extract as unrelated glyphs
## Summary

`try_remap_subset_cmap` applies its sequential-GID repair to any Identity-H/V CID font. That repair assumes CIDs are glyph indices a subsetter can renumber, which holds for `CIDFontType2` (TrueType) but not for `CIDFontType0` (CFF), where CIDs are resolved through the CFF charset. A valid ToUnicode CMap on a CFF font therefore stays valid after subsetting, and renumbering it replaces correct text with unrelated glyphs.

For CFF fonts the corrupting path is unavoidable: the `CIDToGIDMap` branch that would repair the CMap correctly can never be taken, because `CIDToGIDMap` is defined only for CIDFontType2 (PDF 32000-1:2008, 9.7.4.2). So any CIDFontType0 font whose `/W` array starts at a low CID falls straight through into `remap_to_sequential`.

Real-world impact: Japanese PDFs using Adobe-Japan1 CFF fonts (e.g. National Diet Library proceedings) extract as long runs of a single unrelated kanji.

## Reproduction

```bash
curl -o diet.pdf "https://kokkai.ndl.go.jp/minutes/api/v1/detailPDF/img/121205378X00120231201"
# sha256: f0a367135f31f7e40eaf36f2294ab98ff2086627228366489f1d21bd83e3f6e0
cargo run --bin pdf2md -- diet.pdf > out.md
head -c 200 out.md
```

Actual (first line):

```
???? 針針針針針針 針針針針針針 針針針針針針 針針針針針針 針針針針針針 針針針針針針 針針針 針針針 ...
```

Expected — what the document actually says. Cross-checked with pdf.js (`pdfjs-dist` 6.0.227, `getTextContent()` over all pages):

```
一 （八〇） 令和五年十二月一日 （金曜日） 午前九時開議 出席委員 法務委員会 委員長 武部 新君 ...
```

Counting replacement characters over the whole document:

```bash
node -e 'const s=require("fs").readFileSync("out.md","utf8");
console.log("chars="+s.length, "U+FFFD="+(s.match(/�/g)||[]).length)'
# chars=69099 U+FFFD=1233   (1.78%)
```

Note the repeated-glyph runs themselves contain no U+FFFD, so a replacement-character heuristic alone does not catch this.

Not WASM-specific: `@firecrawl/pdf-inspector` 1.11.2 (napi) and `@firecrawl/pdf-inspector-wasm` 0.1.3 both produce the same broken text (identical character and U+FFFD counts to the `main` build).

## Root cause

The offending font (`obj=120`) is:

- `/Subtype /Type0`, `/Encoding /Identity-H`
- descendant `/Subtype /CIDFontType0`, `/CIDSystemInfo` with `/Ordering (Japan1)`
- **no `/CIDToGIDMap`** (correct — it is a CIDFontType2-only entry)
- `/W` array starting at CID 1
- a valid `/ToUnicode` CMap (the file carries 10 of them)

`RUST_LOG=pdf_inspector::tounicode=debug` confirms the remap fires:

```
DEBUG pdf_inspector::tounicode] Subset GID mismatch detected for obj=120:
W starts at CID 1, CMap min CID 633. Remapping to sequential.
```

Path through `src/tounicode.rs`:

1. `try_remap_subset_cmap` — Identity-H, so it proceeds
2. `min_source_cid()` is 633 (> 2), so it proceeds
3. `get_cid_to_gid_map()` returns `None` — CFF fonts have no `CIDToGIDMap`, so the correct repair branch is skipped
4. `get_w_array_start_cid()` returns 1 (≤ 2) → treated as "sequential post-subset GIDs"
5. `remap_to_sequential()` sorts every source CID and reassigns them to 1, 2, 3, … — destroying the correct mapping

## Relation to existing issues (checked before filing)

- **#118** (closed) reports a similar-looking symptom on Type0/Identity-H fonts, but it is a different defect: there the ToUnicode CMaps are *authored* garbled and no decode source can recover the text, so #120 resolved it by flagging the page for OCR (`CipherGarbleStats`) rather than by repairing the CMap. That PDF (`text_simple__att10k.pdf`, sha256 `8b9aecf1…`) uses **CIDFontType2** descendants, so it is unaffected by the fix below — I ran it before and after and the output is byte-identical (3033 chars both ways).
- **66c88a4**, which introduced `remap_to_sequential`, describes the target as generators that "subset-embed fonts by renumbering GIDs sequentially" — a CIDFontType2 scenario by construction. It added only synthetic unit tests; no fixture PDF exercises the path on a CFF font.
- #122 and #202 are adjacent (ToUnicode validity reporting, U+FFFD handling) but neither covers this.

## Proposed fix

Restrict the whole repair block to CIDFontType2 descendants, placed **before** the `CIDToGIDMap` branch so that a `CIDToGIDMap` wrongly attached to a CFF font by a malformed producer is ignored too.

After the guard, the Diet PDF goes from 1233 U+FFFD to 0, and character 3-gram recall against a hand-written ground truth rises from 0.354 to 0.605. `cargo fmt --check`, `cargo clippy -- -D warnings` and `cargo test` (861 tests) all pass.

PR follows. Happy to take the fuller route instead (reading the embedded CFF charset to resolve CID→GID properly) if you would prefer that — it changes considerably more than a guard, so I would rather confirm the direction first.

## Environment

- `pdf-inspector` `main` @ a15ec2d (2026-07-31)
- also reproduced on `@firecrawl/pdf-inspector` 1.11.2 and `@firecrawl/pdf-inspector-wasm` 0.1.3
- macOS 15 (Darwin 25.5.0), aarch64, rustc stable

m-naoki-m:
Fixes #208.

## What

`try_remap_subset_cmap`'s sequential-GID repair assumes CIDs are glyph indices a subsetter can renumber. That holds for `CIDFontType2` (TrueType) but not for `CIDFontType0` (CFF), where CIDs are resolved through the CFF charset — so a valid ToUnicode CMap stays valid after subsetting, and renumbering it corrupts otherwise-correct text.

For CFF fonts the corrupting path was unavoidable: `CIDToGIDMap` is CIDFontType2-only (PDF 32000-1:2008, 9.7.4.2), so the branch that repairs the CMap correctly can never be taken, and any CFF font whose `/W` array starts at a low CID fell through into `remap_to_sequential`.

This guards both repair paths on a CIDFontType2 descendant, placed **before** the `CIDToGIDMap` branch so a `CIDToGIDMap` wrongly attached to a CFF font by a malformed producer is ignored as well.

## Verification

On the National Diet Library proceedings PDF from #208 (sha256 `f0a36713…`):

| | before | after |
|---|---|---|
| U+FFFD | 1233 / 69099 chars (1.78%) | **0 / 68331** |
| repeated-glyph runs | present | gone |
| character 3-gram recall vs. hand-written ground truth | 0.354 | **0.605** |

The recall figure is a character-level 3-gram recall (whitespace and markdown punctuation stripped) against a manual transcription of the document's opening pages. Word-token metrics are not meaningful for Japanese, which has no inter-word spaces.

The PDF behind #118 (`text_simple__att10k.pdf`, sha256 `8b9aecf1…`, CIDFontType2 descendants) extracts **byte-identically** before and after — 3033 chars both ways.

`cargo fmt --check`, `cargo clippy -- -D warnings` and `cargo test` all pass (861 tests: 717 unit + 141 integration + 2 + 1 doc).

## Test changes

Two existing tests build descendant font dicts without a `/Subtype`, so they would stop exercising their intended path once the guard is added. Both set `CIDToGIDMap`, which is CIDFontType2-only, so `Subtype = CIDFontType2` is what the fixtures were already describing:

- `test_try_remap_fires_for_true_subset_mismatch` — would otherwise return early and never reach the remap it asserts on (fails loudly)
- `test_try_remap_skipped_when_w_covers_cmap` — would otherwise still pass, but for the wrong reason, no longer checking the W-coverage logic at all (a silent false positive)

Added `test_try_remap_skipped_for_cid_font_type0`, covering a plain CFF descendant and one carrying a spurious `CIDToGIDMap`.

## Scope note

This narrows an existing heuristic to the font type it was written for; it does not make that heuristic more precise. Separately: `/W` describes glyph *widths*, and a CID absent from `/W` legitimately falls back to `/DW`, so "`/W` starts low and does not cover the CMap's max CID" is suggestive of subset renumbering rather than proof of it — a compliant CIDFontType2 font with a sparse `/W` plus `/DW` could still trip it. That is pre-existing behaviour and orthogonal to this fix; happy to open a separate issue if you consider it worth tightening.

## Not verified

`pdf-evals` is not publicly accessible, so I could not run the 187-PDF snapshot suite or `bench.py score` that `AGENTS.md` asks for. Everything else in the checklist passes locally. Would appreciate a maintainer running that suite before merge.

## Remaining gap (not addressed here)

After the fix the Diet document still scores below pdf.js on character 3-gram recall (0.605 vs 0.780). The residual is a reading-order/layout question in this document class (mixed vertical/horizontal minutes), not an encoding one — the glyphs themselves now decode correctly. Happy to file that separately.

## Alternative considered

Reading the embedded CFF charset to resolve CID→GID properly would be the fuller fix, but it changes considerably more than a guard. Restricting the existing heuristic to the font type it was written for is the minimal correct change. Let me know if you would prefer the charset route and I will rework it.

---
<a href="https://app.blacksmith.sh/firecrawl/codesmith/pdf-inspector/pr/209"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788259527&installation_model_id=11126&pr_number=209&repository=firecrawl%2Fpdf-inspector&return_to=https%3A%2F%2Fgithub.com%2Ffirecrawl%2Fpdf-inspector%2Fpull%2F209&signature=46aa3f560b670a34323912236bd39832af9f2e3330909be8953f55cd3b946bba"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Skip subset GID remapping for `CIDFontType0` (CFF) descendants to avoid corrupting valid ToUnicode mappings. Resolve indirect `/Subtype` and only skip when the descendant is explicitly not `CIDFontType2`; any `CIDToGIDMap` on CFF is ignored.

- **Bug Fixes**
  - Guard both repair paths on `CIDFontType2`, resolving indirect `/Subtype`; missing/unresolvable `/Subtype` keeps prior behavior.
  - Place the guard before the `CIDToGIDMap` branch; ignore `CIDToGIDMap` on CFF.
  - Results: NDL proceedings PDF now has 0 U+FFFD and higher 3-gram recall (0.605 vs 0.354); a `CIDFontType2` sample extracts byte-identically.
  - Tests: reverted earlier fixture edits; CFF regression uses a real `CIDToGIDMap` stream; added an indirect `/Subtype` test.

<sup>Written for commit eeb696f017695b7a5e07530ad71492be1bbdf5b2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/pdf-inspector/pull/209?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/tounicode.rs
+++ b/src/tounicode.rs
@@ -880,6 +880,22 @@ fn try_remap_subset_cmap(
         None => return (cmap, None),
     };
 
+    // Both repair paths below assume CIDs are glyph indices that a subsetter can
+    // renumber, which is only true for CIDFontType2 (TrueType). For CIDFontType0
+    // (CFF), CIDs are resolved through the CFF charset, so a valid CMap stays valid
+    // after subsetting and renumbering it corrupts otherwise-correct text.
+    // CIDToGIDMap is likewise CIDFontType2-only (PDF 32000-1:2008, 9.7.4.2), so this
+    // also ignores a CIDToGIDMap that a malformed producer attached to a CFF font.
+    let is_cid_font_type2 = cid_font_dict
+        .get(b"Subtype")
+        .ok()
+        .and_then(|o| o.as_name().ok())
+        .is_some_and(|n| n == b"CIDFontType2");
+    if !is_cid_font_type2 {
+        debug!("Subset remap skipped for obj={obj_num}: not a CIDFontType2 descendant");
+        return (cmap, None);
+    }
+
     // If there's an explicit CIDToGIDMap, build a repaired CMap using it.
     if let Some(cid_to_gid) = get_cid_to_gid_map(cid_font_dict, doc) {
         if let Some(repaired) = build_cmap_with_cid_to_gid_map(&cmap, &cid_to_gid) {
@@ -3093,6 +3109,7 @@ endbfrange
         // Build a CIDFont dict with Identity CIDToGIDMap and a W array that
         // covers CID 633 via `597 [widths...]`.
         let mut cid_font = lopdf::Dictionary::new();
+        cid_font.set("Subtype", lopdf::Object::Name(b"CIDFontType2".to_vec()));
         cid_font.set("CIDToGIDMap", lopdf::Object::Name(b"Identity".to_vec()));
         cid_font.set(
             "W",
@@ -3136,6 +3153,7 @@ endbfrange
 
         let mut doc = Document::new();
         let mut cid_font = lopdf::Dictionary::new();
+        cid_font.set("Subtype", lopdf::Object::Name(b"CIDFontType2".to_vec()));
         cid_font.set("CIDToGIDMap", lopdf::Object::Name(b"Identity".to_vec()));
         cid_font.set(
             "W",
@@ -3158,5 +3176,53 @@ endbfrange
             remapped.is_some(),
             "Remap must fire when CMap's CIDs are outside W array coverage"
         );
+    }
+
+    #[test]
+    fn test_try_remap_skipped_for_cid_font_type0() {
+        // Same W/CMap mismatch as the CIDFontType2 case above, but the descendant is
+        // CIDFontType0 (CFF). There CIDs are resolved through the CFF charset, so the
+        // ToUnicode CIDs stay valid after subsetting and must not be renumbered.
+        // Real-world case: Japanese Adobe-Japan1 PDFs (e.g. National Diet Library
+        // minutes) where remapping turned correct text into unrelated glyphs.
+        let cmap_content = r#"
+1 begincodespacerange
+<0000><FFFF>
+endcodespacerange
+1 beginbfrange
+<0200> <0220> <0410>
+endbfrange
+"#;
+        let cmap = ToUnicodeCMap::parse(cmap_content.as_bytes()).unwrap();
+
+        let mut doc = Document::new();
+        let mut cid_font = lopdf::Dictionary::new();
+        cid_font.set("Subtype", lopdf::Object::Name(b"CIDFontType0".to_vec()));
+        // CIDToGIDMap is CIDFontType2-only, but a malformed producer can still emit
+        // it on a CFF font. It must not be used to rewrite the CMap either.
+        cid_font.set("CIDToGIDMap", lopdf::Object::Name(b"Identity".to_vec()));
+        cid_font.set(
+            "W",
+            lopdf::Object::Array(vec![
+                lopdf::Object::Integer(1),
+                lopdf::Object::Array(vec![lopdf::Object::Integer(500); 34]),
+            ]),
+        );
+        let cid_font_id = doc.add_object(cid_font);
+
+        let mut font_dict = lopdf::Dictionary::new();
+        font_dict.set("Encoding", lopdf::Object::Name(b"Identity-H".to_vec()));
+        font_dict.set(
+            "DescendantFonts",
+            lopdf::Object::Array(vec![lopdf::Object::Reference(cid_font_id)]),
+        );
+
+        let (primary, remapped) = try_remap_subset_cmap(cmap, &font_dict, &doc, 789);
+        assert!(
+            remapped.is_none(),
+            "Remap must be skipped for CIDFontType0 (CFF) descendants"
+        );
+        // The original CMap must still resolve its own CIDs.
+        assert_eq!(primary.lookup(0x0200), Some("\u{0410}".to_string()));
     }
 }
--- a/src/tounicode.rs
+++ b/src/tounicode.rs
@@ -886,13 +886,16 @@ fn try_remap_subset_cmap(
     // after subsetting and renumbering it corrupts otherwise-correct text.
     // CIDToGIDMap is likewise CIDFontType2-only (PDF 32000-1:2008, 9.7.4.2), so this
     // also ignores a CIDToGIDMap that a malformed producer attached to a CFF font.
-    let is_cid_font_type2 = cid_font_dict
-        .get(b"Subtype")
-        .ok()
-        .and_then(|o| o.as_name().ok())
-        .is_some_and(|n| n == b"CIDFontType2");
-    if !is_cid_font_type2 {
-        debug!("Subset remap skipped for obj={obj_num}: not a CIDFontType2 descendant");
+    // /Subtype may be an indirect reference, so resolve it through the document.
+    // Only bail out when the descendant is *explicitly* something other than
+    // CIDFontType2: a missing or unresolvable /Subtype keeps the previous
+    // behaviour rather than silently disabling the repair.
+    let subtype = cid_font_dict.get(b"Subtype").ok().and_then(|o| match o {
+        Object::Reference(r) => doc.get_object(*r).ok().and_then(|o| o.as_name().ok()),
+        other => other.as_name().ok(),
+    });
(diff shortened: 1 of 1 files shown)


m-naoki-m:
Thanks — both valid. Fixed in eeb696f.

**1. Indirect `/Subtype` (P2)** — agreed, and this was the more serious of the two: a genuine CIDFontType2 font storing `/Subtype` indirectly would have been read as "not CIDFontType2", returned early, and lost the repair — reintroducing exactly the corruption this PR fixes, for those fonts. The guard now resolves the reference through the document.

I also took the second half of the suggestion: it now bails out only when `/Subtype` is *explicitly* a non-CIDFontType2 name, so a missing or unresolvable `/Subtype` keeps the pre-existing behaviour rather than silently disabling the repair.

That had a welcome side effect. The two existing tests only needed their fixtures touched because a missing `/Subtype` used to fall on the skip side; now they pass unmodified, so **the diff against `main` is additive only** (121 insertions, 0 deletions) and no existing test is altered.

**2. CIDToGIDMap coverage in the CFF test (P3)** — correct, `/Identity` is treated as "no map" by `get_cid_to_gid_map`, so the test could not detect the guard being moved. It now attaches a real `CIDToGIDMap` stream (cid 1 → gid 0x0200, which the CMap resolves). I verified this by moving the guard back below the `CIDToGIDMap` branch and confirming `test_try_remap_skipped_for_cid_font_type0` fails.

Also added `test_try_remap_resolves_indirect_subtype` for the first point.

`cargo fmt --check`, `cargo clippy -- -D warnings` and `cargo test` (862 tests) pass. The Diet PDF from #208 still extracts with 0 U+FFFD, and the #118 PDF is still byte-identical to before the change.

abimaelmartell:
hey @m-naoki-m, thanks for the contribution!

m-naoki-m:
Thanks for merging!

PR #221, opened 2026-08-02
fix: decompression-bomb OOM guard + cubic-cost bailout in chart-cluster detection

MADENIYOU:
## Summary

Two issues found while profiling this crate against real and adversarial PDFs, each fixed in its own commit:

1. **Decompression-bomb OOM** — `lopdf::Stream::decompressed_content()` inflates FlateDecode streams fully into memory with no size cap. A ~5MB crafted PDF (single content stream, ~400:1 compression ratio) drove `detect-pdf`'s peak RSS past **2GB** in under 3 seconds. This happens before any of this crate's own guards (e.g. `MAX_OPERATIONS`) run, since those bound *parsed* content, not the raw bytes produced by inflating it.
2. **Cubic-cost table/chart detection** — `is_chart_bar_cluster` (used to distinguish bar charts from cell-rect grids) does an O(n³)-ish nested scan over a rect cluster. A real fixture already in this repo's test suite, `tests/fixtures/bits_pilani_feedback.pdf` (a dense feedback-form PDF with hundreds of touching rects per page), took **43s** to process against the ~200ms this crate documents — profiling showed this one function consuming ~25% of total runtime.

## Fix 1: bounded decompression (`src/safe_decompress.rs`, new)

- `decompressed_content_capped` re-implements the plain-FlateDecode path with a 64MiB output cap (`flate2::read::ZlibDecoder` + `Read::take`), falling back to `lopdf`'s own decoder for cases the cap can't safely cover: uncompressed streams (can't expand), non-Flate filters (LZWDecode/ASCII85Decode — not the observed vector), and PNG/TIFF-predictor streams (predictor post-processing is image/xref territory, not content streams; deferring avoids risking wrong bytes for a case this crate doesn't exercise).
- `get_page_content_capped` mirrors `Document::get_page_content` (the main extraction entry point) using the capped decoder, since that method does its own unbounded decompression internally and isn't something we can patch from outside `lopdf`.
- Every production call site of `.decompressed_content()` in `src/` (detector, content-stream extraction, Form XObjects, embedded font programs, CMap/ToUnicode parsing) now goes through the capped path. `src/bin/dump_ops.rs`, a local debugging tool never fed untrusted input, is left as-is to keep the diff scoped to the actual attack surface.

**Verified**: `detect-pdf` peak RSS on the crafted PDF drops from ~2.03GB to ~99MB; runtime from 2.83s to 0.49s.

## Fix 2: bound `is_chart_bar_cluster`'s worst case (`src/tables/detect_rects.rs`)

- Adds `MAX_CHART_CLUSTER_RECTS = 300` (mirroring the existing `MAX_CLUSTER_RECTS` pattern already in this file for the same reason) as an early bailout: clusters above that size are treated as "not chart-like" instead of running the expensive nested checks. No real bar chart has hundreds of bars, so this costs no detections against realistic PDFs while keeping the worst case bounded — the case that made the cost explode is a dense form/checkbox grid, not a legitimate chart.
- Hoists the `it.page == page` filter out of `numeric_or_empty` (previously re-checked per rect on every call) since it's invariant for the whole `is_chart_bar_cluster` invocation.
- I intentionally avoided a deeper algorithmic rewrite of the O(n³) logic itself — this heuristic has a documented history of behavior regressions against a private eval suite (`pdf-evals`, referenced in this file's comments and `CLAUDE.md`) that I don't have access to, so the safest change is one that can't alter output on any realistic input, only on pathological ones.

**Verified**: the fixture drops from 43s to 4.5s (~9.5x). Added a regression test (`is_chart_bar_cluster_bails_out_above_max_size`) asserting the bailout fires in <100ms on an oversized synthetic cluster.

## Testing

- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` (only pre-existing, unrelated `useless_vec` lint noise in `src/tables/grid.rs` remains — confirmed present on `main` before this branch too, likely a stricter lint in a newer local clippy than whatever's pinned in CI; not touched here to keep the diff scoped)
- `cargo test`: 721 tests pass (720 existing + 1 new), 0 failures
- Manual verification of both fixes with before/after timing and `/usr/bin/time -l` memory measurements (numbers above)

## How I found these

Profiled `pdf2md`/`detect-pdf` with `samply` + macOS `sample` against this repo's own largest fixture, and separately crafted a minimal single-stream decompression-bomb PDF with Python's `zlib` to test memory behavior on adversarial input, since this crate is explicitly designed to process PDFs from untrusted sources.

---
<a href="https://app.blacksmith.sh/firecrawl/codesmith/pdf-inspector/pr/221"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788285539&installation_model_id=11126&pr_number=221&repository=firecrawl%2Fpdf-inspector&return_to=https%3A%2F%2Fgithub.com%2Ffirecrawl%2Fpdf-inspector%2Fpull%2F221&signature=d5bcfc608c7612e8fc69b7668faf3ed3fbc8404101158290e4d7cc357642aae2"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Adds bounded decompression for PDF streams (64 MiB per stream, 64 MiB per page's combined content) so crafted decompression-bomb PDFs can't exhaust memory, and makes extraction fail closed with a typed `PdfError::ResourceLimit` instead of silently returning partial results. The chart-cluster detection optimization originally in this PR was split out after a table-detection regression, so this PR is focused on the decompression fix.

- **Bug Fixes**
  - `safe_decompress::decompressed_content_capped` caps `FlateDecode`/`LZWDecode` and `ASCII85Decode` (with checked groups and a bound on `z`), applies PNG/TIFF predictors after bounded inflate, and includes raw-deflate recovery; `decompressed_or_raw` skips over-cap streams and only falls back to raw on genuine decode errors.
  - Page content uses lopdf 0.44's native `get_page_content_with_limit` but returns `PdfError::ResourceLimit` on a cap hit instead of silently returning empty content.
  - `process_document` now propagates `PdfError::ResourceLimit` from Mixed PDFs instead of swallowing it in the non-fatal extraction fallback.
  - Adds fixtures and integration tests for oversized content streams, split content at token boundaries, and Mixed PDFs with oversized streams; the Python bindings suite now expects these fixtures to raise a resource-limit `ValueError`.

- **Dependencies**
  - Added `flate2` and `weezl` for bounded decompression of `lopdf` streams.

<sup>Written for commit 8f02c0f7ccf15ba022ea5f93b60ebd6749f8086d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/pdf-inspector/pull/221?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MADENIYOU:
Addressed all 7 findings from the automated review. Pushed as two follow-up commits:

**P0 — predictor bypass (safe_decompress.rs:57)**: Confirmed and fixed. Predictor-bearing FlateDecode streams were falling through to `lopdf`'s unbounded decoder, reopening the OOM. Now the cap is enforced on the raw inflate first (predictor post-processing doesn't change output length, so applying it after capping is safe) via `lopdf::filters::png::decode_frame` (public). Re-verified with a predictor-bearing bomb PDF: peak RSS stays at ~104MB instead of climbing back to ~2GB.

**P1 — LZWDecode bypass (tounicode.rs:2325)**: Added a bounded LZW decoder using `weezl`'s incremental `decode_bytes` API, checking cumulative output against the cap after every chunk, configured to match lopdf's exact decoder setup.

**P1 — aggregate page cap (safe_decompress.rs:88)**: Added `MAX_PAGE_CONTENT_BYTES` (256 MiB) enforced while concatenating a page's content streams in `get_page_content_capped`.

**P2 — silent garbage fallback (content_stream.rs:233)**: Agreed this was a real issue, and it also applied to every other call site using the same `unwrap_or_else(|_| stream.content.clone())` pattern, not just this one. Split the error type into `Failed` (genuine decode error — falls back to raw bytes, matching lopdf's own behavior) vs `ExceedsCap` (cap hit — now skips the stream instead of feeding still-compressed bytes into whatever parses the result). Added `decompressed_or_raw()` so this distinction applies uniformly everywhere instead of re-deciding it per call site.

**P2 — lost raw-deflate recovery (detector.rs:761)**: Replicated lopdf's exact corrupt-zlib-header recovery (retry as raw deflate, skip the 2-byte header) inside the bounded path, so streams that were previously recoverable still are.

**P2 — chart-cluster false negatives (detect_rects.rs:2024)**: Fair point — 300 was tighter than necessary. Instrumented the actual pathological fixture to measure real cluster sizes: the problematic clusters run ~2000-2008 rects (near the existing `MAX_CLUSTER_RECTS` ceiling), so raised the cap to 500 — an order of magnitude of headroom below the pathological range while giving real margin above plausible legitimate multi-series/stacked charts. Runtime on the fixture is ~5s at 500 vs ~4.5s at 300, both a large improvement over the original 43s.

**P3 — flaky timing assertion (detect_rects.rs:4713)**: Dropped the wall-clock assertion from the regression test; it now only checks correctness (oversized cluster isn't misclassified as a chart). The perf claim is verified manually against the real fixture instead.

Also: ASCII85Decode got a small local decoder (output can only shrink, so no cap needed) so multi-filter chains like `[ASCII85Decode, FlateDecode]` stay on the bounded path end-to-end rather than falling back to `lopdf`'s unbounded decoder for the whole chain — this wasn't separately called out but follows from the same "any filter chain touching an expanding filter must be fully bounded" principle as the P0 finding.

New tests cover PNG predictor application, LZW round-trip (encoded with `weezl`'s own encoder), ASCII85 (including the `z` shorthand and a chained-filter case), and the unimplemented-filter error path. Full suite: 727 tests passing, `cargo clippy --all-targets -- -D warnings` introduces no new findings versus `main` (28 pre-existing/unrelated errors in both, from a newer local clippy than whatever's pinned in CI).

MADENIYOU:
Addressed both findings from the follow-up review:

**P3 — ASCII85 overflow (safe_decompress.rs:304)**: Confirmed valid. \`decode_ascii85\` was using wrapping arithmetic, so a malformed group encoding a value past \`u32::MAX\` (up to \`85^5 - 1\`, which the digit range technically allows) would silently wrap into an arbitrary decoded value instead of erroring. Switched to checked arithmetic — an overflowing group now returns \`DecompressError::Failed\`, routing through the same raw/skip recovery as every other malformed stream. Added a test with the maximal group (\`"uuuuu"\`) confirming rejection.

**P3 — weak regression test (detect_rects.rs:4709)**: Also valid, and a good catch — the oversized cluster was geometrically degenerate (touching rects, single y, no items), so it was already failing \`bar_family\`'s own checks independent of the size guard; the test would keep silently passing even if the guard were removed. Added \`chart_shaped_cluster()\`, building genuinely chart-like bars (spaced apart, strictly increasing heights, numeric labels), plus a sibling test proving that exact shape returns \`true\` at n=10 (well under the cap). The oversized test now uses the same shape at n=501, so its \`false\` result can only come from the size bailout — exactly as suggested.

Full suite: 729 tests passing, clippy stays at the same 28 pre-existing/unrelated baseline errors (one new type-complexity lint on the test helper's tuple return addressed with a scoped \`#[allow]\`).

abimaelmartell:
@cubic-dev-ai

MADENIYOU:
All five confirmed and fixed.

**P1 — ASCII85 'z' shorthand bypasses the cap**: Right — 'z' is the decoder's only sub-linear-input expansion path (1 byte in, 4 zero bytes out) and had no bound of its own. Now checks the remaining budget before each expansion; added \`ascii85_z_shorthand_respects_the_decompressed_size_cap\` (a stream of enough 'z's to exceed the cap correctly returns \`ExceedsCap\`).

**P2 — aggregate cap not a real memory cap**: Confirmed, and worth detailing: the real mechanism was subtler than "truncate leaves capacity allocated" — appending each stream's data via \`extend_from_slice\` in full, then separately \`push\`ing the join newline, let the newline push alone trigger \`Vec\`'s amortized-doubling growth policy whenever \`content\` was exactly at capacity, compounding across streams (5×55MiB streams ended up retaining ~461MiB, not ~256MiB). Fixed by appending only up to the remaining budget and sizing a single \`reserve_exact\` call to cover both the data *and* the separator, so no implicit growth policy ever runs. New test \`get_page_content_capped_never_exceeds_aggregate_cap\` asserts both length and retained capacity stay bounded — verified it fails against the original code (capacity ~461MiB) and passes with the fix.

**P2 — missing separator between streams**: Confirmed. Added a fixture (\`split_content_stream_no_separator.pdf\`) where stream 1 ends exactly at \`Tj\` and stream 2 begins exactly at \`ET\` — without a joining newline these merge into the invalid token \`TjET\` and the shown text is dropped. New test \`split_content_stream_without_separator_does_not_drop_text\`; verified it fails without the newline join and passes with it.

**P2 — oversized chart cluster misroutes to table detection**: Confirmed, good catch. \`is_chart_bar_cluster\` returning \`false\` for anything over \`MAX_CHART_CLUSTER_RECTS\` meant a genuinely huge chart wouldn't be excluded — it'd fall through to table-grid detection and render as a garbage table. Since \`has_chart_bar_signature\` is roughly cubic in cluster size, running it on the full oversized cluster isn't an option, so instead it now runs on an evenly-strided sample bounded to the cap. A real bar chart's repeated structure (spacing, height variation) survives subsampling. New test \`is_chart_bar_cluster_detects_genuine_chart_above_max_size\` (chart-shaped geometry past the cap is still detected as a chart) plus \`is_chart_bar_cluster_still_bounded_and_correct_on_oversized_dense_grid\` (a dense uniform checkbox-style grid — the actual pathological case the cap exists for — still correctly resolves as not-a-chart).

**P3 — doc/PR description says 300, code says 500**: The shipped constant is 500 (intentionally tuned, per the earlier review round) — the discrepancy was in the description, not the code. No code change needed there.

Full suite: 895 tests passing (6 new), clippy clean on all three changed files.

MADENIYOU:
All three confirmed and fixed.

**P2 — sampling over raw storage order, not spatial order**: Confirmed. Sorted by x before striding, so the sample is spatially even regardless of how the cluster's rects happen to be stored. Honest note on the test: I tried to construct a fixture that concretely demonstrates the *old* code failing (a "two series stored series-by-series" reordering of the chart-shaped geometry), but \`bar_family\`'s spacing/height-variation signal turned out robust enough that even that adversarial reordering still passed against the un-sorted code — I couldn't produce a compact repro that actually fails without the fix. Shipping the sort anyway since it's strictly safer (still O(n log n), no risk of reintroducing the cubic cost the cap exists to avoid) and directly matches what you flagged; the new test (\`is_chart_bar_cluster_detects_genuine_chart_above_max_size_out_of_spatial_order\`) verifies correctness on non-spatially-ordered input even though it doesn't prove a prior regression.

**P3 — stale ASCII85 doc comments**: Confirmed — those comments predated the 'z' cap fix and were flatly contradicted by it. Updated both the module-level and function-level comments to describe the expansion and why the cap has to be enforced during decoding, not just after. Also moved a misplaced doc comment that was sitting above \`decode_ascii85_group\` but actually described \`decode_ascii85\` (looks like a pre-existing placement slip, unrelated to this PR's changes) to the function it belongs to.

**P3 — decompressing before checking the exhausted budget**: Confirmed, straightforward reorder — moved the \`remaining == 0\` check above the \`decompressed_content_capped\` call, matching the truncation break below it.

Full suite: 896 tests passing (1 new), clippy clean on both changed files.

MADENIYOU:
Confirmed — you're right, the sort-then-stride sample from the last round fixed spatial representativeness but at the cost of exactly what you flagged: an input-sized copy and O(n log n) sort, which isn't truly bounded against an adversarial cluster size.

Replaced with a single O(n) pass and O(cap) extra memory: find the cluster's x-range in one pass, then keep one representative rect per bucket across \`MAX_CHART_CLUSTER_RECTS\` evenly-spaced x buckets — no full-size copy, no sort. Still depends only on each rect's own x value (not storage order), so the out-of-spatial-order test from the last round still passes unchanged.

Worth being upfront: this is still O(n) *time* (one pass over the input is unavoidable if the sample has to depend on the actual rect positions at all — a blind O(1) bailout can't be "representative" of anything), but it no longer allocates or sorts proportionally to the input, which is the part your comment specifically called out. Also surfaced an unrelated limitation while stress-testing at extreme scale: bucketing by x-value can under-represent a chart whose real signal is concentrated in a narrow x-sub-range relative to the cluster's full extent (verified this doesn't happen for realistic chart geometry — only for a degenerate 200k-rect test construction whose height formula grows unbounded, itself hitting an unrelated pre-existing bar-width sanity filter). Noted in the new test's comments rather than silently working around it.

New test \`is_chart_bar_cluster_sampling_does_not_allocate_input_sized_buffers\`, verifying correctness (not misclassifying a dense grid as a chart) at 200,000 rects with no strict timing assertion (this repo has a documented history of dropping flaky timing assertions) — the whole suite still running in ~4s is itself decent evidence the O(n)/O(cap) approach stays cheap. Full suite: 897 tests passing (1 new), clippy clean.

MADENIYOU:
Confirmed and fixed — good catch, and I should have caught this myself given `has_chart_bar_signature` explicitly checks both orientations.

Extracted the bucketing into a shared `bucket_by_axis()` helper and now bucket by both x and y (each into half the cap), taking the union. A horizontal bar chart's common left edge (near-constant x) no longer collapses into one bucket — the y-axis bucketing keeps its row signal, while the x-axis bucketing still handles the vertical-bar case. Still O(n) time, O(cap) memory — no full-size copy or sort.

New test `is_chart_bar_cluster_detects_genuine_horizontal_chart_above_max_size` with a mirrored `horizontal_chart_shaped_cluster` helper (common x=0 left edge, increasing width, varying y). Verified it fails when temporarily reverted to x-only bucketing and passes with the fix.

Full suite: 898 tests passing (1 new), clippy clean.

MADENIYOU:
Confirmed and fixed. Halving the cap between the two axes meant neither orientation got the full resolution it might need — a multi-series vertical chart with many bars packed into a wide x-range could need close to the full cap's worth of x-buckets to keep each series distinct, while the y-axis pass (added for horizontal bars) contributes nothing useful when y is near-constant, as it usually is for a standard vertical chart's baseline.

Fixed by giving each axis the full cap rather than splitting it. The union is still a hard, cluster-size-independent bound — at most 2x \`MAX_CHART_CLUSTER_RECTS\`, a fixed constant regardless of how large the adversarial input is — just with full resolution per orientation instead of half.

Honest note: I tried constructing a compact fixture that concretely demonstrates a multi-series vertical chart failing at half-cap resolution and passing at full-cap, but couldn't build one without introducing confounding factors (denser packing than the half-cap-vs-full-cap distinction alone starts tripping unrelated touching-rect/background-fill guards elsewhere in the heuristic). The existing chart-detection tests (including the horizontal and out-of-order ones from the last two rounds) all still pass, and the fix itself is a direct, low-risk implementation of what you suggested — but I don't have a test that specifically isolates this one.

Full suite: 898 tests passing, clippy clean.

rmacy:
I rebased this PR onto current `main` and tested the public Rust and Python paths. The decompression cap works, but the cap-exceeded path can still return partial extraction as success.

`decompressed_or_raw` converts `ExceedsCap` into an empty byte vector. Page content, Form XObjects, and Form ToUnicode streams can therefore disappear without a typed error.

I wrote the regression tests first and observed these failures:

- a page content stream above 64 MiB did not return a public resource-limit error
- a Form XObject stream above 64 MiB did not return a public resource-limit error
- content after an exactly exhausted 256 MiB page budget could be skipped

My local follow-up:

- adds `PdfError::ResourceLimit { page, object_id, resource, limit_bytes }`
- propagates the error through page content, Form XObject, and Form ToUnicode extraction
- exposes `PdfResourceLimitError` through Python
- preserves the exact 256 MiB boundary and rejects the next byte

Validation on the combined branch:

- `cargo fmt --check`
- `cargo clippy --offline -- -D warnings`
- `cargo test --all-targets --offline` — 1,020 passed
- `cargo build --release --offline`
- 18-fixture, 132-page local torture corpus — 90/90 pdf-inspector runs completed within the process limits

No external service handled the PDF bytes. I recommend adding the typed, fail-closed path to this PR. I can also open it as a stacked follow-up if that is easier to review.

MADENIYOU:
@rmacy Thanks — confirmed, and thanks for the exact repro shape. Implemented the core of this (page content extraction) directly on this PR rather than waiting on your diff, since the fix is well-contained here.

Added `PdfError::ResourceLimit { page, object_id, resource, limit_bytes }` and changed `get_page_content_capped` to return `Result<Vec<u8>, PdfError>`, failing closed as soon as either the per-stream (64 MiB) or aggregate (256 MiB) cap is hit, instead of silently skipping/truncating. Propagated through `extract_page_text_items` via `?` — it already returned `Result<_, PdfError>`, so this was the main page-content path and didn't need a broader signature refactor.

Verified with a fixture (two `/Contents` streams: one small stream with genuine text so cheap detection classifies the page as TextBased and extraction actually runs, one that decompresses to ~70 MiB) and confirmed it fails without the fix — `process_pdf_with_options` previously returned `Ok(PdfProcessResult { markdown: Some("## Real visible text on this page\n"), .. })`, silently dropping the oversized stream's content with no signal anything was wrong. That's the exact bug you described.

**Scope I didn't cover**: Form XObject and ToUnicode streams (`extractor/xobjects.rs`, `tounicode.rs`) still go through `decompressed_or_raw` and don't surface a typed error yet. Those call sites are inside a recursive, non-`Result`-returning helper (`extract_form_xobject_text_inner`, depth-tracked nested form recursion), and converting them raises a real design question I don't think I should decide unilaterally: should one oversized *nested* XObject fail the whole page, or just skip that XObject's contribution and keep the rest (matching how multiple page-content streams already degrade independently before this fix)? Given you've already got this designed and validated (1,020 tests, the 18-fixture torture corpus), I'd rather you open that as a follow-up — happy to review it, and I think you're better positioned to make that call given the validation work you've already done.

Full suite: 900 tests passing (4 new), clippy clean on all four changed files.

MADENIYOU:
Both confirmed and fixed.

**P1 — Mixed-type fallback swallows ResourceLimit**: Confirmed. The \`extracted.ok()\` fallback was designed for soft extraction failures (garbled text, etc.) where falling back to "no markdown, but still a valid result" makes sense — it wasn't updated to treat the new \`ResourceLimit\` variant as the hard, fail-closed case it's meant to be. Now checked and propagated before the Mixed-type \`.ok()\` conversion.

New fixture (real text + a large declared-only image, so the page classifies as \`Mixed\`, plus an oversized third content stream) and integration test, verified to fail without the fix — it previously returned \`Ok(PdfProcessResult { pdf_type: Mixed, markdown: None, .. })\` with no error surfaced at all, silently indistinguishable from a page that genuinely had no text.

**P3 — undocumented variant**: Added an enum-level doc comment on \`PdfError\` distinguishing \`ResourceLimit\` (valid PDF, but hit the decompression-bomb guard) from the other, input-validation-style variants, so it's discoverable via \`cargo doc\` and callers know to match on it explicitly if they need to.

Full suite: 902 tests passing (2 new), clippy clean.

abimaelmartell:
I verified this against current `main` and a synthetic merge of the PR.

The decompression issue is real and the guard works. The included 72 KB fixture has a 71 KB stream that inflates to 70 MiB (~1,028x). Current `main` accepts it, returns partial Markdown, and peaked around 302 MB RSS locally; this PR fails closed with `PdfError::ResourceLimit` at 64 MiB. The synthetic merge also passed `cargo fmt -- --check`, `cargo clippy -- -D warnings`, the full test suite (1,050 unit + 168 integration, plus binaries/docs), and release builds.

I do not think we should merge the current combined scope, though. A quick regression run found a statistical-table PDF where `main` produces the correct 20-column table, while this PR produces a malformed 9-column table (92.7% snapshot similarity). Restoring only `src/tables/detect_rects.rs` to `main` made that output byte-identical again while preserving the decompression protection, so the regression is isolated to the bundled chart-cluster optimization rather than the security fix.

Could you please drop/split the `detect_rects.rs` chart work from this PR and keep this PR focused on bounded decompression and the resource-limit behavior? We can revisit the performance change separately with the regression fixture in the eval set.

One remaining scope note: oversized nested Form XObject and ToUnicode streams still pass through `decompressed_or_raw` and become silent skips rather than propagating `ResourceLimit`. I am okay treating that as an explicit follow-up so the P0 page-content protection can land quickly, but it should not get lost.

Once the PR is narrowed, we can run the full eval suite. If you cannot get to the scope cleanup in the next few days, let us know and we can take over the rework.

MADENIYOU:
@abimaelmartell Thanks for the precise regression fixture and diagnosis — confirmed and done.

**Scope narrowed**: rebased this branch onto current `main` (it had drifted quite a bit — lopdf 0.42 → 0.44, #478's independent decompression fix, and a lot of unrelated feature work) and restored `src/tables/detect_rects.rs` to `main`'s exact version, dropping the chart-cluster bucket-sampling optimization entirely. Diff against `main` on that file is now empty. Agreed that work should be revisited separately with your regression fixture in the eval set — happy to pick it back up properly scoped once you've got that in place, or leave it to whoever picks it up.

**One thing the rebase surfaced**: `main` had independently landed its own page-content decompression bound in the interim (#478, using lopdf 0.44's native `get_page_content_with_limit`) — the exact same call site this PR was patching with a hand-rolled equivalent. Rather than keep two overlapping implementations, I kept `main`'s native call (no reason to carry a redundant custom reimplementation of protection lopdf now provides upstream) and changed its error path to fail closed with `PdfError::ResourceLimit` instead of `main`'s current silent empty-result skip — so the fail-closed guarantee this PR set out to add is preserved on top of the now-canonical bound. Removed the now-dead `get_page_content_capped` wrapper and its unreachable aggregate cap along with its 3 now-untestable unit tests; the 2 ResourceLimit integration tests (real fixtures, end to end) already cover this behavior against the current implementation.

**Scope note carried forward**: your remaining flag on nested Form XObject / ToUnicode streams still passing through `decompressed_or_raw` as silent skips (not `ResourceLimit`) stands as-is — still explicitly out of scope here, as you suggested.

Verified the merged branch's `cargo clippy --all-targets -- -D warnings` output is byte-identical (same errors, same files) to running clippy on a clean `main` checkout side by side — zero new findings from this change. Full suite: 1174 unit + 179 integration + 2 doc tests, all green, including the snapshot suite that would have caught the chart-cluster regression.

Ready for the eval suite whenever convenient.

MADENIYOU:
CI catch: the "OCR runtime smoke" job's \`TestMultipleFixtures.test_process_all_fixtures\` parametrizes over every \`.pdf\` in \`tests/fixtures/\` and asserts \`process_pdf\` succeeds — my two new fixtures (\`oversized_content_stream.pdf\`, \`mixed_oversized_content_stream.pdf\`) are deliberately malicious and are *supposed* to make it raise, so the test correctly flagged the mismatch:

\`\`\`
ValueError: resource limit exceeded on page 1, object (4, 0): page content exceeds 67108864 bytes
\`\`\`

Fixed by adding a targeted exception list (mirroring the existing \`ENCRYPTED_FIXTURE_PASSWORDS\` pattern already in that file) — these two fixtures now assert the expected \`ValueError\` is raised, rather than being silently skipped or failing the catch-all.

Verified locally with \`maturin develop --release --features python\` + pytest: fails without the fix (reproduces the exact CI failure) and passes with it. Full local Python suite: 97 passed.

PR #224, opened 2026-08-02
perf: enable LTO and codegen-units=1 for release builds

### LINKED ISSUE #95: Enable Link-Time Optimization (LTO) and codegen-units = 1 for Release builds (and possibly other options)
Hi!

I noticed that in the `Cargo.toml` file [Link-Time Optimization](https://nnethercote.github.io/perf-book/build-configuration.html#link-time-optimization) (LTO) for the project is not enabled. I suggest switching it on since it will reduce the binary size (always a good thing to have) and will likely improve CPU performance a bit due to more aggressive compiler optimizations. Especially since you already [enabled](https://github.com/firecrawl/pdf-inspector/blob/8b63ceb084f75dfd4425129d1d90d5bca45054e4/napi/Cargo.toml#L17) for the Napi part.  Additionally, `codegen-units = 1` (CU1) option can help too in a similar to LTO way, so I recommend to enable it as well - including the Napi part too.

I recommend enabling LTO only for Release builds so developers experience won't be affected by the increased build time. Actually, I can propose to use flags directly from this `ripgrep` [profile](https://github.com/BurntSushi/ripgrep/blob/cbc598f245f3c157a872b69102653e2e349b6d92/Cargo.toml#L76) (like stripping and other potentially useful for the project switches).

Basically, it can be enabled with the following lines to the root Cargo.toml file:

```
[profile.release]
codegen-units = 1
lto = true # FatLTO - the most aggressive LTO version
<possible other options like strip = true>
```

I have made quick tests (AMD Ryzen 9 5900x, Fedora 44, Rust 1.95, `cargo build --release` and `maturin build --release` build commands) - here are the results:

| Binary name \ Build Profile | Current Release | Release + FatLTO + CU1 |
| --- | --- | --- |
| `pdf2md` | 6.3 Mib | 5.0 Mib | 
| `detect-pdf` | 6.3 Mib | 5.0 Mib |
| `libpdf_inspector.so` | 6.5 Mib | 5.1 Mib |
| Python Wheel | 2.4 Mib | 2.1 Mib |

Clean build time (`cargo build --release`):

* Current Release: 13s
* Release + FatLTO + CU1: 57s

Clean build time (`maturin build --release`):

* Current Release: 11s
* Release + FatLTO + CU1: 47s

Build time increase shouldn't be a problem since we enable it only for the Release profile - in this case, we would not affect the development lifecycle. If the proposed settings are added to the Cargo.toml file, all binaries will be built with the new options automatically.

Thank you.

MADENIYOU:
Fixes #95.

## Summary

`napi/` and `wasm/` already set `lto = true` in their own `[profile.release]` (`wasm/` also already had `codegen-units = 1`), but the root crate — used for the Rust library, the `pdf2md`/`detect-pdf` CLI binaries, and the Python wheel via `maturin` — had no `[profile.release]` section at all, so it built with Cargo's defaults (`lto = false`, `codegen-units = 16`).

## Change

Added the same profile the other two bindings already use:

```toml
[profile.release]
lto = true
codegen-units = 1
strip = "debuginfo"
```

`strip = "debuginfo"` wasn't explicitly requested in the issue but matches `napi/`'s existing profile, so debug builds stay untouched and all three release profiles are now consistent.

## Verification

| | Before | After |
|---|---|---|
| `pdf2md` binary size | 6,592,512 bytes | 5,448,176 bytes (~17% smaller) |
| `detect-pdf` binary size | 6,592,512 bytes | 5,448,176 bytes (~17% smaller) |

Full test suite (859 tests — lib, bins, doctests) passes unchanged in release mode. Release build time goes from ~11s to ~1m40s on this machine, as expected for LTO + a single codegen unit — this only affects `cargo build --release`, not the day-to-day debug-build dev loop.

---
<a href="https://app.blacksmith.sh/firecrawl/codesmith/pdf-inspector/pr/224"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788289963&installation_model_id=11126&pr_number=224&repository=firecrawl%2Fpdf-inspector&return_to=https%3A%2F%2Fgithub.com%2Ffirecrawl%2Fpdf-inspector%2Fpull%2F224&signature=338a6dab0a39a900f79ee17520e2330912c638a360aea608214ba369b9af9844"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Enable LTO and `codegen-units = 1` in the root `[profile.release]` to produce smaller, faster binaries, consistent with `napi/` and `wasm/`. Also sets `strip = "debuginfo"`; `pdf2md`/`detect-pdf` are ~17% smaller and release builds take longer, while debug builds are unchanged.

<sup>Written for commit 3b60055fed17ea3ec4bb029e9cb09969d8870aa2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/pdf-inspector/pull/224?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,6 +59,14 @@ include_dir = "0.7"
 [dev-dependencies]
 tempfile = "3.3"
 
+# Matches the release profile already used in napi/ and wasm/: LTO and a
+# single codegen unit trade build time for a smaller, faster binary. Debug
+# builds are untouched, so the local dev loop isn't affected.
+[profile.release]
+lto = true
+codegen-units = 1
+strip = "debuginfo"
+
 [features]
 default = []
 python = ["pyo3"]


MADENIYOU:
Small correction on the framing, then the substance: I didn't claim this keeps debug info in release builds — the PR description says "debug **builds** stay untouched," meaning \`cargo build\` (the non-\`--release\` dev profile), not release-build debug info.

That said, the underlying point is real and worth stating plainly: \`strip = "debuginfo"\` does strip DWARF symbols from release binaries, so a release-only crash or a profiler like \`samply\`/\`perf\` attached to \`cargo build --release\` output will show addresses instead of function names and source locations, until symbols are supplied separately (e.g. via \`splitdebuginfo\` or a separate \`dsym\`/\`.debug\` artifact, which this PR doesn't set up).

I included it mainly for consistency — \`napi/\`'s own \`[profile.release]\` already makes this exact tradeoff, and this PR just extends it to the root crate rather than introducing a new precedent. But given it's genuinely a debuggability-vs-size tradeoff rather than a strict improvement, happy to drop the line (or switch to \`strip = "none"\`) if you'd rather keep symbols in release binaries — it's a one-line change either way. Let me know which you'd prefer and I'll push it.

PR #226, opened 2026-08-02
fix: don't drop inline subscripts/superscripts near page margins as page numbers

MADENIYOU:
## Summary

A report on this repo's recent activity said pdf-inspector "seems to drop superscripts and subscripts silently." Investigated and confirmed a real bug, isolated to a specific mechanism — not a vague/general problem with sub/superscript handling.

## Root cause

\`is_page_number\` in \`src/extractor/layout.rs\` flags any 1-4-digit item positioned near the top/bottom margin (\`y > 720 || y < 100\` on a page) as a page number, with **no check for whether the item is actually isolated on its own line**.

A subscript or superscript digit that fails to merge into its parent word — \`merge_text_items\` requires font sizes within 20% of each other, which a subscript/superscript by definition isn't — stays a separate, short, numeric \`TextItem\`. When that item happens to sit near a page's top or bottom margin (titles, headers, and footnote references routinely do — a document title with a formula, a chemistry abstract, a footnote marker), it gets swept up by this filter and **vanishes from markdown output entirely** — not merged into its parent, not left as a bare digit, just gone.

## Minimal repro

A one-page PDF: "H" (14pt) + "2" (9pt, subscript-positioned) + "O" (14pt), near the top of the page.

| | Before | After |
|---|---|---|
| \`pdf2md --raw\` | \`H O\` | \`H2O\` |
| \`pdf2md --items-json\` | "2" present (confirms extraction itself is fine — the loss is specifically in this filter) | unchanged |

## Fix

Only drop a page-number candidate when nothing else on the page is **tightly X-adjacent** to it at a similar Y (near-touching — gap less than 1.5× the larger of the two items' font sizes). Y-proximity alone isn't a safe signal: a genuine page number commonly shares its footer row with an unrelated element (a tagline, a running title) at ordinary word-gap distance, and that combination must still be filtered.

This precision requirement was caught empirically, not assumed — a first pass that checked only Y-proximity regressed the existing \`nexo-price-en.pdf\` snapshot test, which has exactly this shape: a page number ("50") sharing a footer row with a marketing tagline ("THE ALL-NEW NEXO /// ECO-FRIENDLY CAR") at ~17pt of separation. The tightened X-adjacency check (gap < 1.5× font_size) correctly keeps dropping that page number while still protecting the H2O case (near-zero gap).

## Bonus: recovers real content in an existing fixture

\`p1244-1996.pdf\`'s title line was rendering as "Form Employee's Report" — the form number **"4070"** (a real IRS form number: [Form 4070, Employee's Report of Tips to Employer](https://www.irs.gov/pub/irs-pdf/f4070.pdf), tightly adjacent to "Form" at the top margin) was being silently dropped by this exact bug. Updated that snapshot to include the recovered text — a content-completeness fix, not a behavior change requiring separate review.

## Testing

Added two fixtures:
- \`inline_subscript_near_margin.pdf\` (the H2O case) — the digit must survive markdown conversion
- \`isolated_page_number_footer.pdf\` (a genuinely lone footer page number, nothing tightly adjacent) — must still be filtered, guards the fix's precision against the nexo-style false positive

\`cargo test\`: 866 tests pass (2 new + existing suite, including both snapshot tests). \`cargo clippy --all-targets -- -D warnings\`: unchanged at 28 pre-existing/unrelated errors versus \`main\`.

---
<a href="https://app.blacksmith.sh/firecrawl/codesmith/pdf-inspector/pr/226"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788304546&installation_model_id=11126&pr_number=226&repository=firecrawl%2Fpdf-inspector&return_to=https%3A%2F%2Fgithub.com%2Ffirecrawl%2Fpdf-inspector%2Fpull%2F226&signature=90535c9eae88dff72012cdd0c1f011697d19781eae9386d49b66a96284d3d901"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Fixes false-positive page-number filtering near margins so inline subscripts/superscripts aren’t dropped. Keeps real footers filtered, including decorated and split-digit numbers; reverts the line‑membership fallback to avoid 3‑column footer leaks.

- **Bug Fixes**
  - Only drop a margin number if it has no tightly X-adjacent alphanumeric, non-page-number neighbor on the same line (gap < 0.25× max font size; y within 5pt). Page-number-shaped neighbors and decorative-only glyphs never protect, preventing split-digit and adorned-footer leakage.
  - Added fixtures/tests for inline sub/superscripts, split-digit numbers, word-adjacent and decorated footers, and a 3‑column footer; reverted the p1244 “Form 4070” snapshot along with the line-membership rule.

<sup>Written for commit 1fe666d9cd85cdd4165301cd80d07f389490252d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/pdf-inspector/pull/226?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/extractor/layout.rs
+++ b/src/extractor/layout.rs
@@ -1264,12 +1264,20 @@ fn group_into_lines_with_thresholds_and_regions_impl(
         // numbers would need a different signal (column-layout awareness,
         // sentence-level semantics) than adjacency; out of scope here.
         //
-        // The neighbor must also not itself be page-number-shaped.
-        // Otherwise a multi-digit page number rendered as separate
-        // single-digit Tj operators (a real PDF-generator pattern) would
-        // have each digit "protect" its sibling — every digit sees a
-        // page-number-like neighbor and none of them get dropped, leaking
-        // the whole number into markdown.
+        // Two more guards on what counts as a protecting neighbor:
+        //
+        // - It must contain at least one alphanumeric character — not just
+        //   be non-empty. Decorative flanking glyphs ("• 12 •", "- 12 -")
+        //   are real, non-empty TextItems sitting tightly adjacent to a
+        //   footer number, but they aren't *content*; protecting the
+        //   number because a bullet happens to sit next to it would leak
+        //   exactly the footer pattern this filter exists to catch.
+        // - It must not itself be page-number-shaped. Otherwise a
+        //   multi-digit page number rendered as separate single-digit Tj
+        //   operators (a real PDF-generator pattern) would have each digit
+        //   "protect" its sibling — every digit sees a page-number-like
+        //   neighbor and none of them get dropped, leaking the whole
+        //   number into markdown.
         let mut items_by_page: HashMap<u32, Vec<&TextItem>> = HashMap::new();
         for item in &items {
             items_by_page.entry(item.page).or_default().push(item);
@@ -1287,7 +1295,7 @@ fn group_into_lines_with_thresholds_and_regions_impl(
                     .unwrap_or(&[]);
                 !page_items.iter().any(|other| {
                     if std::ptr::eq(*other, item)
-                        || other.text.trim().is_empty()
+                        || !other.text.chars().any(|c| c.is_alphanumeric())
                         || is_page_number(other)
                         || (other.y - item.y).abs() >= y_tol
                     {
--- a/src/extractor/layout.rs
+++ b/src/extractor/layout.rs
@@ -1237,9 +1237,59 @@ fn group_into_lines_with_thresholds_and_regions_impl(
     // Markdown output omits standalone numeric headers/footers. Plain-text
     // callers opt out because dropping extracted text violates that API.
     let items = if filter_page_numbers {
+        // `is_page_number` only looks at one item's own text/position — it
+        // can't tell a lone footer digit from a short numeric item that
+        // happens to sit inline next to other text (a subscript/
+        // superscript that failed to merge into its parent, positioned
+        // near the top/bottom margin: a title-line formula, an exponent,
+        // a footnote reference). Only drop a page-number candidate when
+        // it's actually isolated — no *tightly adjacent* neighbor on the
+        // same line.
+        //
+        // The neighbor check requires near-touching X proximity, not just
+        // shared Y: a real page number commonly shares its footer row with
+        // an unrelated element (a marketing tagline, a document title) at
+        // ordinary word-gap distance, and that must still be dropped. A
+        // subscript/superscript that failed to merge into its parent sits
+        // right against it (near-zero gap) — that's the actual signal to
+        // protect, not "anything else on the page at a similar height".
+        let mut items_by_page: HashMap<u32, Vec<&TextItem>> = HashMap::new();
+        for item in &items {
+            items_by_page.entry(item.page).or_default().push(item);
+        }
+        let y_tol = 5.0;
+        let drop: Vec<bool> = items
+            .iter()
+            .map(|item| {
+                if !is_page_number(item) {
+                    return false;
+                }
+                let page_items = items_by_page
+                    .get(&item.page)
+                    .map(Vec::as_slice)
+                    .unwrap_or(&[]);
+                !page_items.iter().any(|other| {
+                    if std::ptr::eq(*other, item)
+                        || other.text.trim().is_empty()
+                        || (other.y - item.y).abs() >= y_tol
+                    {
+                        return false;
+                    }
+                    let gap = if item.x <= other.x {
+                        other.x - (item.x + item.width.max(0.0))
+                    } else {
+                        item.x - (other.x + other.width.max(0.0))
+                    };
+                    let max_font_size = item.font_size.max(other.font_size);
+                    gap < max_font_size * 1.5 && gap > -max_font_size
+                })
+            })
+            .collect();
         items
             .into_iter()
-            .filter(|item| !is_page_number(item))
+            .zip(drop)
+            .filter(|(_, drop)| !*drop)
+            .map(|(item, _)| item)
             .collect()
     } else {
         items
--- a/src/extractor/layout.rs
+++ b/src/extractor/layout.rs
@@ -1243,16 +1243,28 @@ fn group_into_lines_with_thresholds_and_regions_impl(
         // superscript that failed to merge into its parent, positioned
         // near the top/bottom margin: a title-line formula, an exponent,
         // a footnote reference). Only drop a page-number candidate when
-        // it's actually isolated — no *tightly adjacent* neighbor on the
-        // same line.
+        // it's actually isolated — no *tightly adjacent* real-text
+        // neighbor on the same line.
         //
-        // The neighbor check requires near-touching X proximity, not just
-        // shared Y: a real page number commonly shares its footer row with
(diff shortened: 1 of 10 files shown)


MADENIYOU:
Both confirmed and fixed, and both caught with real fixtures rather than just tightening a number blind.

**P2 — split-digit self-protection (layout.rs:1272)**: Fixed by excluding \`is_page_number(other)\` from the protector check — a digit can no longer be "protected" by a neighboring digit that's itself page-number-shaped. Added \`split_digit_page_number_footer.pdf\` ("12" rendered as two separate single-digit \`Tj\`s): previously both digits leaked through, now correctly dropped as a whole.

**P2 — threshold too loose (layout.rs:1284)**: Confirmed — \`1.5x font_size\` is well past ordinary word-gap spacing (~0.28x for a space character in a typical proportional font), so it was retaining real page numbers next to unrelated text at normal spacing. Tightened to \`0.25x\`, calibrated just above \`merge_subscript_items\`'s own \`0.2x\` merge-gap tolerance — tight enough to reject a normal inter-word space, loose enough to still protect a subscript/superscript that just barely misses that merge for a structural (non-gap) reason. Added \`word_adjacent_page_number_footer.pdf\` ("12" at normal word-gap distance from "Company Name", different font size so it survives \`merge_text_items\` as its own item): previously retained, now correctly dropped.

**Side effect worth flagging directly**: rebuilding the fixtures with accurate font metrics (a real \`/Widths\` array, instead of relying on pdf-inspector's width=0 fallback for an unmetriced font) exposed that the original "H2O" fixture only passed by coincidence — the inaccurate width math inflated its gap, but it still happened to land under the old loose threshold. With real widths, that specific case's true gap is near-zero and already merges via \`merge_subscript_items\` on its own (verified separately, no fix needed there). The new fixture instead targets this PR's actual scope: a numeric subscript/superscript that \`merge_subscript_items\` **structurally** declines to merge — a parent ending in a digit rather than a letter, or a strikeout/underline mismatch — regardless of gap tightness.

One more honest tradeoff: this tightening reverts the "Form 4070" bonus recovery mentioned in the original PR description. That case's real gap (~0.94-1.35x font_size) is a stylistic size-emphasis gap in a title, not a failed-merge subscript's near-zero one — there's no single threshold that protects it without also re-admitting the false positives this review caught. Reverted the p1244-1996.md snapshot line; recovering that case would need a different signal than X-adjacency and is out of scope here.

Full suite: 866 tests passing (4 new), clippy unchanged at the same 28 pre-existing/unrelated baseline errors.

MADENIYOU:
Both fixed.

**P3 — over-broad assertion**: Scoped to whole-word matches, matching the sibling tests. Also found the fixture itself wasn't testing what it claimed: the two digits shared a font size, so \`merge_text_items\` fused "1"+"2" into one "12" item *before* \`is_page_number\` ever saw two separate candidates — the self-protection bug wasn't actually being exercised. Rebuilt with differing font sizes so they genuinely stay separate through the filter.

**P2 — you're right, the snapshot revert masked the regression instead of fixing it.** A form number embedded in a real title losing its content is exactly the class of bug this PR is about, just via a different code path than the original subscript case.

Fixed properly this time with an independent signal that doesn't depend on gap-size tuning at all: **line membership**. If a page-number candidate shares its Y with at least two *other* real-text items, that's a genuine multi-word sentence/title line — not a footer row, which pairs a page number with at most one decorative element (a tagline, a company name). This sidesteps the whole "how tight is tight enough" tension entirely, since a title number's gap to its neighbors (widened by visual size emphasis) doesn't need to be confused with footer-adjacent spacing — footers just don't have two independent real-text neighbors to begin with.

Verified against the exact case you flagged: p1244-1996's "Form 4070 Employee's Report" line has three items ("Form", "4070", "Employee's Report") sharing that Y — now protected. Re-checked everything from the first review pass stays correctly filtered (nexo's footer, the split-digit and word-adjacent fixtures — all two-items-total lines). Added \`title_embedded_number.pdf\`, a synthetic fixture mirroring the real shape, with a dedicated test. Restored the p1244-1996.md snapshot to recover "4070" again, this time backed by a real mechanism instead of an accidental threshold.

Full suite: 868 tests passing (1 new), clippy unchanged at the same 28 pre-existing/unrelated baseline.

MADENIYOU:
Valid catch, and it exposed something worth stating plainly rather than patching around again.

Built the exact fixture your comment described — a 3-column footer ("Acme Corporation | 12 | Annual Report 2026", left label / centered page number / right label). Its column-stop gaps (~107-131pt) turned out to **overlap the exact range** the real p1244 title-line gap needs to be protected ("Form"→"4070" ≈9.4pt, "4070"→"Employee's Report" ≈107pt). I tried tightening the contiguity threshold, but any value generous enough to protect the real title case also protects this synthetic footer's page number — they're not separable by gap size or line-membership alone, no matter where the line is drawn.

Rather than keep tuning a threshold that provably can't distinguish the two cases, I reverted the line-membership fallback entirely. The fix is now scoped to just tight X-adjacency (0.25x font_size) — the mechanism that's been solid across all three review passes and fully resolves the confirmed original bug (subscripts/superscripts near page margins silently dropped). Recovering "Form 4070"-style title numbers would need a genuinely different signal (column-layout awareness, sentence-level semantics), not adjacency — flagging that as out of scope here rather than shipping something that trades one false-positive class for another.

Replaced \`title_embedded_number.pdf\` with \`three_column_footer.pdf\` as a permanent regression test, so this exact mechanism can't get silently reintroduced later. Reverted the p1244-1996.md snapshot's "4070" recovery to match.

Full suite: 868 tests passing, clippy unchanged at the same 28 pre-existing/unrelated baseline.

MADENIYOU:
Confirmed and fixed. Decorative flanking glyphs ("• 12 •", "- 12 -") are real, non-empty TextItems, so they satisfied the old "non-empty" protector check even though they're not content — leaking that common decorated-footer pattern.

Changed the check from "non-empty" to "contains at least one alphanumeric character." Added a fixture ("12" flanked by "." decorations in a different font size, so they survive `merge_text_items` as separate items and genuinely reach `is_page_number` as tightly-adjacent-but-non-alphanumeric neighbors): leaked through before, correctly dropped after.

Full suite: 870 tests passing (1 new), clippy unchanged at the same 28 pre-existing/unrelated baseline.

PR #230, opened 2026-08-03
fix: recover from a corrupted startxref pointer

### LINKED ISSUE #228: No xref recovery: files that pypdf and pdfium repair raise Invalid PDF structure
## Summary

PDFs with a damaged cross-reference offset are entirely unprocessable — every entry point raises — while `pypdf` and `pdfium` both detect the damage and recover by rebuilding the xref from a scan of the file body.

In a corpus of ~100 real-world business documents, **5 (~5%)** failed this way. All 5 parse without incident in both other engines.

## Reproduction

Fully self-contained (`pip install pdf-inspector pypdf reportlab`). It writes a valid one-page PDF, then corrupts the `startxref` offset to point at the wrong byte — which is what damaged writers emit in the wild:

```python
import io

# 1. A perfectly ordinary one-page PDF.
c = canvas.Canvas("valid.pdf", pagesize=letter)
c.setFont("Helvetica", 10)
c.drawString(72, 720, "Order Detail Report by Account")
c.drawString(72, 700, "03/14/2024   IC-1048   WIDGET ASSEMBLY   Shipped   $482,110.40")
c.showPage()
c.save()

# 2. Point startxref at a wrong-but-plausible offset.
raw = open("valid.pdf", "rb").read()
idx = raw.rfind(b"startxref")
broken = raw[: idx + 10] + b"9" + raw[idx + 11:]
open("broken_startxref.pdf", "wb").write(broken)

# 3. Compare engines.
import pdf_inspector
from pypdf import PdfReader

data = open("broken_startxref.pdf", "rb").read()

try:
    pdf_inspector.classify_pdf_bytes(data)
    print("pdf-inspector: OK")
except Exception as e:
    print("pdf-inspector:", type(e).__name__, e)

print("pypdf:", len(PdfReader(io.BytesIO(data)).pages), "pages")
```

### Actual output

```
pdf-inspector: ValueError Invalid PDF structure
incorrect startxref pointer(2)
parsing for Object Streams
pypdf: 1 pages
```

`pypdf` logs that the pointer is wrong, rebuilds, and returns the page. `pdfium` (via liteparse) also parses it silently. `pdf-inspector` raises.

Real-world files in our corpus produced two variants of the failure:

```
ValueError: Invalid PDF structure
ValueError: PDF parsing error: couldn't parse input: invalid file trailer
```

Both raised from `classify_pdf_bytes()` and `extract_pages_markdown_bytes()` alike. Two of the five affected files were ordinary 9-page text documents with roughly 97,000 characters of perfectly extractable text — nothing exotic about them beyond the malformed offset.

### Expected

A PDF whose object data is intact but whose xref offset is wrong should be recoverable, as it is in the two most widely used engines.

## Why it matters

These aren't adversarial or truncated files — they're routine output from software that writes a slightly-off xref, and they're common enough to be ~5% of a real corpus.

The bigger problem is that they fail at *classification* too, not just extraction. That means there's no way to detect the condition and route around it from within the library: a caller needs a second, independent PDF engine just to discover whether pdf-inspector can read a given file at all. That undercuts the "classify cheaply, then decide" workflow the README describes.

## Suggested fix

Add xref reconstruction: on trailer/startxref parse failure, scan the file for `obj` markers and rebuild the object table, as pypdf and pdfium do. Gating it behind an opt-in (`recover=True`) with strict parsing as the default would be entirely reasonable.

Failing that, a distinguishable exception type for "malformed but possibly recoverable" would at least let callers fall back deliberately, rather than catching a generic `ValueError` that could mean anything.

## Environment

- Compared against `pypdf==6.14.2` and pdfium (via `liteparse==2.10.1`)
- Python 3.13
- macOS 26.6, arm64

MADENIYOU:
Fixes #228.

## Summary

A PDF whose `startxref` pointer has been corrupted to point at the wrong byte offset — a single flipped digit, which is what damaged writers emit in the wild — was entirely unprocessable: every entry point (`classify_pdf`, `extract_pages_markdown`, `process_pdf`) raised `Invalid PDF structure`, even though the file's object data, real xref table, and trailer were all completely intact just past the wrong pointer. Both `pypdf` and `pdfium` recover from this by locating the real table directly instead of trusting the pointer; `lopdf` doesn't.

## Fix

Added a new repair candidate alongside the existing missing-`%%EOF`-marker and stripped-leading-bytes repairs already in `repair_pdf_container_candidates`: scan the buffer for the real, standalone `xref` keyword and append a corrected trailing `startxref`/`%%EOF` block. `lopdf`'s own `get_xref_start` always reads the *last* `%%EOF` in the final 512 bytes of the buffer and the `startxref` value immediately before it, so the appended block transparently supersedes the corrupted one already in the file — no in-place byte surgery on content the original writer produced.

Scoped to classic (non-stream) xref tables, matching the reported repro and the common case. A corrupted pointer into a cross-reference *stream* (`N 0 obj << /Type /XRef ...>>`, used by some PDF 1.5+ writers instead) would need the containing object's number, not just a byte offset — out of scope here, happy to follow up separately if useful.

## Verification

Reproduced the issue's exact repro (a valid one-page PDF, then corrupt the `startxref` offset to point at the wrong byte):

| | Before | After |
|---|---|---|
| `process_pdf` / `classify_pdf` / `extract_pages_markdown` | `Invalid PDF structure` on all three | Loads correctly |
| Extracted text | n/a (unprocessable) | "Order Detail Report by Account", "WIDGET ASSEMBLY", the dollar amount — all present |
| Page count | n/a | 1 (correct) |

Added a regression test using a fixture built the same way as the issue's repro script.

## Testing

- `cargo test`: 859 tests pass (1 new)
- `cargo clippy --all-targets -- -D warnings`: unchanged at 28 pre-existing/unrelated errors versus `main`

---
<a href="https://app.blacksmith.sh/firecrawl/codesmith/pdf-inspector/pr/230"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788348983&installation_model_id=11126&pr_number=230&repository=firecrawl%2Fpdf-inspector&return_to=https%3A%2F%2Fgithub.com%2Ffirecrawl%2Fpdf-inspector%2Fpull%2F230&signature=5d4747d739fbbef65cf8f5875170205406b30c952ff8aa6340c515625e4da8c8"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Recover PDFs with a corrupted `startxref` by scanning for the real classic `xref` table and appending a correct trailer. The scan validates the header shape and count termination, runs in a single reverse pass, and fixes “Invalid PDF structure” errors in `process_pdf`, `classify_pdf`, and `extract_pages_markdown`.

- **Bug Fixes**
  - Add a repair path that finds a standalone classic `xref` with a valid `<start-id> <count>` header and appends `startxref`/`%%EOF`, ensuring `lopdf` uses the corrected trailer.
  - Reject false matches and malformed headers (e.g., substring in `startxref`, headerless `xref`, or `xref\n0 6garbage`) to avoid bogus offsets.
  - Use a single reverse scan (O(n)) for better worst‑case performance.
  - Applied to both the original and stripped buffers. Added unit and regression tests. Classic tables only (no xref streams). Behavior matches `pypdf`/`pdfium`.

<sup>Written for commit dc2f79b0a494abecf21e0ee8ee8667788f5484b5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/pdf-inspector/pull/230?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MADENIYOU:
Both confirmed and fixed.

**P2 — trusting any standalone "xref" unconditionally**: Added \`looks_like_xref_subsection_header\`, which confirms a plausible classic xref subsection header (\`<start-id> <count>\`, e.g. "0 6" — the shape every real table starts with) actually follows the candidate token before accepting it. The scan now walks backward until it finds a token that both stands alone *and* validates, instead of accepting the first (rightmost) standalone match unconditionally.

**P2 — quadratic worst case**: Rewrote as a single reverse byte-index walk — O(n) regardless of how many false candidates get rejected along the way, instead of re-invoking `windows().rposition()` on a shrinking prefix per failed candidate.

Added direct unit tests on the byte-level scan rather than trying to construct an adversarial full PDF for the coincidental-match case — that scenario can't actually be represented in a PDF-level integration test anyway, since reportlab (and most real writers) compress page content by default, so a literal "xref" word in visible text never appears as raw bytes in the file. Tests cover: a coincidental standalone "xref" with no subsection header is rejected; a real classic table is found; a coincidental match positioned *after* the real table doesn't shadow it; "xref" as a substring of "startxref" still doesn't match.

Full suite: 863 tests passing (5 new), clippy unchanged at the same 28 pre-existing/unrelated baseline. The original #228 repro is unaffected — reverified manually alongside the existing integration test.

MADENIYOU:
Confirmed and fixed in the latest commit.

**P2 — count run not bounded**: `looks_like_xref_subsection_header` checked that a digit run followed the whitespace separator, but never verified where that run ended. A coincidental `xref\n0 6garbage` in stream/literal content had the right shape (digits, whitespace, digits) and would validate, letting the recovery repair against a bogus offset. Now requires the count run to end at whitespace or buffer-end.

Added `find_xref_rejects_count_run_with_trailing_garbage` covering exactly that shape. Full suite: 869 tests passing (1 new), clippy clean on the changed file (pre-existing unrelated failures remain in `src/tables/grid.rs`, untouched by this PR).

PR #231, opened 2026-08-03
fix: extract_pages_markdown's needs_ocr now agrees with classify_pdf

### LINKED ISSUE #227: needs_ocr from extract_pages_markdown() contradicts classify_pdf() on the same page
## Summary

The per-page OCR verdict is reachable through two public APIs, and they can return **opposite answers for the same page**. `classify_pdf()` appears correct; the `needs_ocr` field on `extract_pages_markdown()` disagrees, and it disagrees in the direction of `False` — i.e. toward silently skipping OCR on a page that needs it.

This shows up on pages that are visually scanned but carry a small amount of genuine native text over the raster — a header line, a stamp, a cover-sheet annotation.

## Reproduction

Fully self-contained (`pip install pdf-inspector reportlab pillow`). It builds a page-sized raster "scan" with no text operators, draws one line of real native text on top, and asks pdf-inspector the same question two ways:

```python
import io
from PIL import Image, ImageDraw
from reportlab.lib.utils import ImageReader

W, H = letter

# 1. A page-sized RASTER (no text operators) that looks like a scanned report.
img = Image.new("RGB", (int(W * 2), int(H * 2)), "white")
d = ImageDraw.Draw(img)
for i, line in enumerate([
    "ORDER DETAIL REPORT BY ACCOUNT",
    "Account IC-1048     Period 07/2023 - 07/2024",
    "Order Date   Item Code      Description      Status    Unit Cost",
    "03/14/2024   IC-1048        WIDGET ASSEMBLY  Shipped   $482,110.40",
    "05/02/2024   IC-1049        BRACKET SET      Shipped   $0.00",
]):
    d.text((100, 120 + i * 46), line, fill=(20, 20, 20))
buf = io.BytesIO()
img.save(buf, format="PNG")
buf.seek(0)

# 2. Draw ONE line of real native text on top of it.
c = canvas.Canvas("scan_with_native_header.pdf", pagesize=letter)
c.drawImage(ImageReader(buf), 0, 0, width=W, height=H)
c.setFont("Helvetica", 11)
c.drawString(48, H - 40, "Order Detail Report by Account")
c.showPage()
c.save()

# 3. Ask the same question two ways.
import pdf_inspector

data = open("scan_with_native_header.pdf", "rb").read()
cls = pdf_inspector.classify_pdf_bytes(data)
ext = pdf_inspector.extract_pages_markdown_bytes(data)

print("classify_pdf     ->", cls.pdf_type, "pages_needing_ocr:", cls.pages_needing_ocr,
      f"(confidence {cls.confidence:.2f})")
print("extract_pages_md -> page 0 needs_ocr:", ext.pages[0].needs_ocr)
print("extracted markdown:", repr(ext.pages[0].markdown))
```

### Actual output

```
classify_pdf     -> image_based pages_needing_ocr: [0] (confidence 0.80)
extract_pages_md -> page 0 needs_ocr: False
extracted markdown: '## Order Detail Report by Account\n'
```

`classify_pdf` gets it right: the page is `image_based` and page 0 needs OCR. `extract_pages_markdown` extracts the single native header line, emits it as `## …`, and reports `needs_ocr=False` — so a caller routing on that field skips OCR and loses the entire scanned table below it.

### Expected

Both APIs should agree on whether a given page needs OCR.

## How often this happens

Measured across 832 pages of real-world business PDFs (110 documents, mixed native-text and scanned):

- **38 pages (4.8%), spanning 8 of 100 documents**, received contradictory verdicts from the two APIs.
- **36 of the 38** disagreed in the unsafe direction (`classify_pdf` = needs OCR, `extract_pages_markdown` = doesn't).
- Checked against the rendered pages, `classify_pdf` was correct in every case examined.

## Why it matters

`extract_pages_markdown()` is the ergonomic place to read this. The result object is already in hand, and a per-page `needs_ocr` sitting directly beside per-page `markdown` reads like the intended per-page OCR-routing signal that the README describes. A caller doing the obvious thing gets the wrong answer on roughly 5% of pages.

The failure is also silent: you receive well-formed Markdown containing just a heading, plus a flag telling you no OCR was necessary. Nothing raises, and nothing in the result indicates the page was mostly image.

## Suggested fix

Have both paths derive from a single classification pass so they cannot disagree.

If the extraction-path flag intentionally means something narrower — "did this extraction invoke OCR" rather than "does this page need OCR" — then renaming it and documenting the distinction would resolve the ambiguity just as well. Currently both are spelled `needs_ocr`, which strongly implies they answer the same question.

## Environment

- Python 3.13
- macOS 26.6, arm64

MADENIYOU:
Fixes #227.

## Summary

`extract_pages_markdown_mem` computed its per-page `needs_ocr` entirely from text-quality signals (decoding/garble issues, empty markdown, GID fonts, garbage-text ratio) — no awareness of the page's image content at all. So a page that is fundamentally a full-page scan with a little genuine native text drawn over it (a header, a stamp, a cover-sheet annotation) extracts that text cleanly, trips none of the text-quality checks, and reports `needs_ocr=false` — while `classify_pdf`/`detect_pdf_type` correctly see the dominant background image and flag the same page as needing OCR. Two public APIs answering the same question, silently disagreeing, in the unsafe direction (skipping OCR on a page that needs it).

## Fix

Exposed `detector::analyze_page_images` at crate visibility (was private) and call it per page in `extract_pages_markdown_mem`'s loop — reusing the exact same "large background image" signal (>50% page coverage) that already powers `has_template_image` in `classify_pdf`/`detect_pdf_type`, rather than reimplementing image-area detection a second time with its own thresholds that could drift out of sync again.

When it's true, the page is flagged `needs_ocr` (with `OCR_REASON_SCANNED` added to `ocr_reasons_by_page`, matching how the same signal is already reported elsewhere) and its markdown is blanked — exactly like the existing text-quality-triggered `needs_ocr` paths already do. No special-casing added for "cleanly-extracted-but-still-a-scan" text.

## Verification

Reproduced the issue's exact scenario (a full-page raster "scan" with one line of genuine native text drawn over it, built via `reportlab`/`pillow`):

| | Before | After |
|---|---|---|
| `classify_pdf_bytes` | `image_based`, `pages_needing_ocr: [0]` | unchanged |
| `extract_pages_markdown_bytes` page 0 | `needs_ocr=False`, markdown = the header line | `needs_ocr=True`, markdown = `""` |

Confirmed no regression on a normal text-based fixture (`nexo-price-en.pdf`): `needs_ocr` stays `False`, full markdown still returned.

## Testing

- New Rust regression test exercising both `detect_pdf_type_mem` and `extract_pages_markdown_mem` against the same fixture, asserting agreement
- `cargo test`: 860 tests pass (1 new)
- `cargo clippy --all-targets -- -D warnings`: unchanged at 28 pre-existing/unrelated errors versus `main`
- Manual verification via the Python bindings (`maturin develop --release --features python`), matching the issue's own repro almost verbatim

---
<a href="https://app.blacksmith.sh/firecrawl/codesmith/pdf-inspector/pr/231"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788349512&installation_model_id=11126&pr_number=231&repository=firecrawl%2Fpdf-inspector&return_to=https%3A%2F%2Fgithub.com%2Ffirecrawl%2Fpdf-inspector%2Fpull%2F231&signature=3ff7e0c272d152fc55dc1f733a629e38ccffa0e09f31828964c4f65a38d0e441"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Aligns `extract_pages_markdown` per-page `needs_ocr` with `classify_pdf` by using the same template-image gates and adding vector-outlined text routing, so scans with incidental native text and outlined-text pages are OCR’d while real text pages with watermark/letterhead are not. Fixes #227.

- **Bug Fixes**
  - `extract_pages_markdown_mem` now gates template-image pages with the same “looks-like-scan” or low-text rules as `classify_pdf` (using `DetectionConfig::min_text_ops_per_page`), routes vector-outlined text pages, adds `OCR_REASON_SCANNED`/`OCR_REASON_VECTOR_TEXT`, and blanks markdown when OCR is required.
  - Added regression tests for a scan with a native header, a text page with a full-bleed watermark (no OCR), and a vector-outlined text page.

- **Refactors**
  - Combined template-image and vector-text checks into `detector::page_ocr_signals` to compute both in one `analyze_page_content` pass and avoid double work.

<sup>Written for commit a18a9472a185cf12a828cbb003d0983d590bc004. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/pdf-inspector/pull/231?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MADENIYOU:
Confirmed and fixed.

**P1 — raw signal too aggressive**: You're right — using `has_template_image` alone dropped the gate classification applies elsewhere, so a text page with a watermark/letterhead/large figure would get correctly-extracted markdown wrongly discarded and routed to OCR.

Digging into *why* classification agrees this specific fixture (`scan_with_native_header_text.pdf`) needs OCR, it turned out to not be `looks_like_scan` at all: that fixture's `unique_alphanum_chars` is 18 (repeated header/footer words), which is *not* `< 10`, so `looks_like_scan` is actually `false` for it. It's flagged via a separate path — `text_operator_count` (1, oddly low due to how the content stream parses) falls below the `effective_min_ops` floor `pages_with_text` requires for image-bearing pages, routing the whole document to `PdfType::ImageBased` (all pages need OCR unconditionally there).

So mirroring only `looks_like_scan` would have fixed the false-positive but broken the #227 regression test. Added `page_template_image_needs_ocr()` in `detector.rs`, which folds in both real signals: `looks_like_scan` (image_count<=1, text<50, low alphanumeric diversity) OR insufficient real text volume (text_operator_count < 10, mirroring the `effective_min_ops` floor). Verified against both fixtures:
- `scan_with_native_header_text.pdf` (#227 regression): text_ops=1 → still correctly flagged.
- New fixture `text_page_with_watermark_image.pdf` (substantial real body text, text_ops=164, over a full-bleed background image): no longer wrongly blanked.

New test `test_extract_pages_markdown_does_not_ocr_text_page_with_watermark_image`. Full suite: 875 tests passing (1 new), clippy clean on both changed files (pre-existing unrelated clippy failures remain elsewhere in the tree, untouched by this PR).

MADENIYOU:
Both confirmed and fixed.

**P1 — missing vector-text signal**: You're right, and it's a real gap independent of the template-image work — `extract_pages_markdown_mem` never checked `has_vector_text` at all. Added `page_has_vector_text()` and wired it into `needs_ocr` the same way `has_template_image` is. Verified with a new fixture (massive path ops simulating outlined glyphs + a short genuine caption, kept under `has_vector_text`'s `unique_alphanum_chars < 30` gate) — confirmed by temporarily un-wiring the fix: the caption extracts cleanly enough (non-empty, non-garbled) that `needs_ocr` was `false` beforehand, exactly the silent disagreement you flagged.

**P2 — hard-coded 10 doesn't match the per-page threshold**: Correct. The `10` was borrowed from `pages_with_text`'s `effective_min_ops` floor, but that's a whole-document aggregate (only applies when deciding overall `PdfType::ImageBased`/`Scanned`), not the per-page Mixed-routing threshold (`config.min_text_ops_per_page`, default 3) this function actually needs to agree with. Switched to `DetectionConfig::default().min_text_ops_per_page`. Re-verified against the #227 regression fixture (text_ops=1, still well under 3 — still correctly flagged) and the watermark fixture (text_ops=164 — still correctly not flagged).

Worth flagging honestly: full parity with the whole-document classification branch isn't achievable from a per-page function without document-level aggregate context (pages_with_text ratios etc.), which `extract_pages_markdown_mem`'s per-page loop doesn't have. This fix narrows the gate to what a single page's own signals actually agree with — the per-page Mixed-routing decision — rather than chasing an unreachable perfect mirror of the whole-document decision tree.

New test `test_extract_pages_markdown_ocrs_page_with_vector_outlined_text`. Full suite: 877 tests passing (2 new), clippy clean on both changed files.

MADENIYOU:
Confirmed and fixed — good catch, that's a real double-decompress per page.

Merged \`page_template_image_needs_ocr\` and \`page_has_vector_text\` into a single \`page_ocr_signals()\` that calls \`analyze_page_content\` once and derives both booleans from the same \`PageAnalysis\`, matching how \`detect_from_document\` caches its per-page analysis instead of recomputing it per signal.

Re-verified all three OCR-consistency regression tests (#227 native-header-over-scan, watermark-text false-positive, vector-outlined-text) still pass. Full suite: 877 tests passing, clippy clean on both changed files.

abimaelmartell:
Hey @MADENIYOU, thanks for the contribution!

PR #232, opened 2026-08-03
fix: currency-prefixed cells wrongly merged as continuation rows in sparse tables

### LINKED ISSUE #229: Sparse/grouped table rows collapse into a single Markdown row with concatenated cells
## Summary

On tables where rows have differing populated-cell patterns — a group-level row with most cells empty, then detail rows beneath it — Markdown extraction **collapses several physical rows into a single table row**, concatenating each column's values inside one cell.

A uniformly-populated table with the same columns extracts perfectly, so this is specific to the sparse shape rather than to the columns or the content.

## Reproduction

Fully self-contained (`pip install pdf-inspector reportlab`). It builds two PDFs with the same columns, differing only in shape — one sparse, one uniform:

```python

W, H = letter
X = [48, 108, 150, 215, 300, 362, 424, 470, 520, 556]
HEADERS = ["Order Date", "Suffix", "Line Section", "Item Code", "Description",
           "Status", "Unit Cost", "Freight Charge", "Tax", "Total Billed"]
DETAIL = [
    ("WIDGET ASSEMBLY", "$482,110.40", "$0.00", "$9,215.75", "$491,326.15"),
    ("BRACKET SET",     "$0.00",       "$0.00", "$0.00",     "$0.00"),
    ("CONTROL MODULE",  "$31,905.22",  "$4,100.00", "$0.00", "$36,005.22"),
]

# --- sparse: group row with empty cells, interleaved note, then detail rows ---
c = canvas.Canvas("sparse.pdf", pagesize=letter)
c.setFont("Helvetica", 7)
y = H - 90
c.drawString(X[1], y, "Account"); c.drawString(X[8], y, "Adjust")   # two-line header
y -= 10
for xi, h in zip(X, HEADERS):
    c.drawString(xi, y, h)
y -= 16
c.drawString(X[0], y, "03/14/2024")          # group row: most cells empty
c.drawString(X[3], y, "IC-1048")
c.drawString(X[5], y, "Shipped")
y -= 12
c.drawString(X[4], y, "Shipment was delayed in transit and re-routed through the hub.")
y -= 12
for desc, unit, freight, tax, total in DETAIL:
    c.drawString(X[3], y, "IC-1048"); c.drawString(X[4], y, desc); c.drawString(X[5], y, "Shipped")
    c.drawRightString(X[7] - 6, y, unit); c.drawRightString(X[8] - 6, y, freight)
    c.drawRightString(X[9] - 6, y, tax);  c.drawRightString(X[9] + 34, y, total)
    y -= 12
c.showPage(); c.save()

import pdf_inspector
print(pdf_inspector.extract_pages_markdown_bytes(open("sparse.pdf","rb").read()).pages[0].markdown)
```

### Actual output — all four rows collapse into one

```
|Order Date|Suffix|Item Code|Description|Status|Unit Cost|Freight Charge|Tax|Total Billed|
|---|---|---|---|---|---|---|---|---|
|03/14/2024||IC-1048 IC-1048 IC-1048 IC-1048|Shipment was delayed in transit and re-routed through the hub. WIDGET ASSEMBLY BRACKET SET CONTROL MODULE|Shipped Shipped Shipped Shipped|$482,110.40 $0.00 $31,905.22|$0.00 $0.00 $4,100.00|$9,215.75 $0.00 $0.00|$491,326.15 $0.00 $36,005.22|
```

Three distinct line items are now one row. `$482,110.40 $0.00 $31,905.22` is a single cell — three different orders' unit costs concatenated.

### Control — the same columns, uniformly populated, extract correctly

Building the same table with every row fully populated and no interleaved note gives exactly the right answer:

```
|Order Date|Item Code|Description|Status|Unit Cost|Qty|Tax|Ref|
|---|---|---|---|---|---|---|---|
|03/14/2024|IC-1048|WIDGET ASSEMBLY|Shipped|$482,110.40|$0.00|$9,215.75|N/A|
|05/02/2024|IC-1049|BRACKET SET|Shipped|$0.00|$0.00|$0.00|N/A|
|07/19/2024|IC-1050|CONTROL MODULE|Open|$31,905.22|$4,100.00|$0.00|N/A|
```

So the trigger is the sparse/grouped shape plus the interleaved free-text line, not the column layout.

On real documents the same bug appears in a milder form — adjacent numeric columns merging pairwise into one cell (e.g. `$0.00 $16,134.34`) rather than the whole table collapsing — but it's the same sparse-row shape driving it.

## Why it matters

The value of table extraction is knowing which column and which row each value belongs to. A collapsed cell forces the consumer to guess, and multiple readings are equally plausible.

The failure is silent and confident: the output is well-formed Markdown asserting a structure that doesn't match the document. Nothing raises, and row-count validation won't catch it, because the row count looks plausible — there just aren't enough of them.

Since the README benchmarks tables via TEDS, this shape may be under-represented in the current corpus. TEDS penalises a merged-but-present cell considerably less than the practical impact warrants, so a benchmark can look healthy while this case is broken.

## Suggested fix

Seed row boundaries from the text items' y-positions and column boundaries from their x-positions, rather than inferring either from rendered density — the geometry is already available and unambiguous here.

When a boundary is genuinely ambiguous, prefer splitting over merging: an extra empty row or column is trivially recoverable downstream, whereas a merged cell has lost information that can't be reconstructed.

## Note

For what it's worth, liteparse 2.10.1 exhibits a similar collapse on the same file, so this seems to be a genuinely hard shape rather than something the field has solved and this library hasn't.

## Environment

- Python 3.13
- macOS 26.6, arm64

MADENIYOU:
Fixes #229.

## Root cause

`clean_table_cells` in `src/tables/format.rs` decides whether a row is a "continuation" (wrapped overflow text that should merge into the previous row) partly based on `looks_like_data_row`, which requires every character in a cell to be `[0-9.,\- ]`. Currency cells like `"$482,110.40"` fail that check outright because of the leading `$`.

In a sparse/grouped table — a group-header row carrying the order date and item code in column 0, followed by detail rows that leave column 0 blank (row-spanning) — each detail row with dollar amounts was misclassified as `!looks_like_data_row`, which let the generic "classic continuation" rule (empty first cell + non-empty other cells) fire and merge it into the previous row. With 3 detail rows in the repro, all their values ended up concatenated into single cells of one merged row instead of staying as 3 distinct rows.

Reproduced and confirmed via a real fixture built from the issue's description (`tests/fixtures/sparse_grouped_table_rows.pdf`):

```
Before:
|03/14/2024||IC-1048 IC-1048 IC-1048 IC-1048|Shipment was delayed... WIDGET ASSEMBLY BRACKET SET CONTROL MODULE|Shipped Shipped Shipped Shipped|$482,110.40 $0.00 $31,905.22|...

After:
|03/14/2024||IC-1048|Shipment was delayed in transit and re-routed through the hub.|Shipped||||
|||IC-1048|WIDGET ASSEMBLY|Shipped|$482,110.40|$0.00|$9,215.75|$491,326.15|
|||IC-1048|BRACKET SET|Shipped|$0.00|$0.00|$0.00|$0.00|
|||IC-1048|CONTROL MODULE|Shipped|$31,905.22|$4,100.00|$0.00|$36,005.22|
```

## Fix

Strip a leading currency symbol (`$`/`€`/`£`/`¥`) before the numeric-shape check in `looks_like_data_row`'s `numeric_cells` count, and require at least one digit remain (so a bare `"$"` or `"-"` cell doesn't count as numeric). This only affects the numeric-shape detection used to protect real data rows from continuation merging — no other heuristic changed.

## Scope note

The interleaved free-text note row in the repro (`"Shipment was delayed in transit..."`, no item code) still merges into the group-header row above it — that's the existing "classic continuation" rule (empty first cell, prose-shaped content) doing what it's designed to do for genuine wrapped text, and is a separate, more debatable heuristic call not part of this issue's reported symptom (rows collapsing into one with concatenated values). Left untouched to keep this fix narrowly scoped.

## Testing

- New unit test `test_clean_table_cells_currency_data_row_not_merged` in `src/tables/format.rs`, covering the exact grouped-row/currency-cell shape.
- New integration test `sparse_grouped_table_rows_stay_separate` in `tests/integration_tests.rs`, using a real fixture PDF built from the issue's example.
- `cargo fmt`, `cargo test` (all passing, 2 new), `cargo clippy --all-targets -- -D warnings` clean on both changed files (pre-existing unrelated clippy failures remain elsewhere in the tree, e.g. `src/tables/grid.rs`, untouched by this PR).

---
<a href="https://app.blacksmith.sh/firecrawl/codesmith/pdf-inspector/pr/232"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788350771&installation_model_id=11126&pr_number=232&repository=firecrawl%2Fpdf-inspector&return_to=https%3A%2F%2Fgithub.com%2Ffirecrawl%2Fpdf-inspector%2Fpull%2F232&signature=a8509ebfb7c740b27fce010f49a9422d026d93d8227f160fbd5009703a127b7b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Fixes incorrect merging of currency-valued detail rows in sparse/grouped tables by treating `$`, `€`, `£`, and `¥`-prefixed amounts as numeric data so detail rows stay separate. Keeps dash/ellipsis placeholders numeric-shaped to avoid regressions. Fixes #229.

- **Bug Fixes**
  - Strip leading currency symbols before the numeric-shape check and remove the digit requirement so `----` and `...` still count as numeric; prevents detail rows from being merged as continuation text.
  - Tests: unit test now mirrors the exact repro (adds Suffix and Total Billed to avoid unrelated heuristics), a new unit test covers dash/ellipsis placeholders, and an integration test with a fixture ensures grouped detail rows remain separate.

<sup>Written for commit 1ccd8d517b67bed19aabba313ca208681638ce07. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/pdf-inspector/pull/232?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/tables/format.rs
+++ b/src/tables/format.rs
@@ -314,7 +314,6 @@ fn clean_table_cells(cells: &[Vec<String>]) -> (Vec<Vec<String>>, Vec<String>) {
                     && c.chars().all(|ch| {
                         ch.is_ascii_digit() || ch == '.' || ch == '-' || ch == ',' || ch == ' '
                     })
-                    && c.chars().any(|ch| ch.is_ascii_digit())
             })
             .count();
         let looks_like_data_row = non_first_cells.len() >= 2
@@ -676,6 +675,57 @@ mod tests {
         assert_eq!(cleaned[2][5], "$482,110.40");
         assert_eq!(cleaned[3][5], "$0.00");
         assert_eq!(cleaned[4][5], "$31,905.22");
+    }
+
+    #[test]
+    fn test_clean_table_cells_dash_and_ellipsis_placeholders_still_count_as_numeric() {
+        // Dash/ellipsis placeholders ("----", "...") in numeric-shaped
+        // columns (common for "not applicable"/zero in financial and
+        // statistical tables) must still count as numeric-shaped cells,
+        // same as before the #229 currency fix — the fix only needed to
+        // strip a leading currency symbol, not require a digit be present.
+        // Same row layout as the currency test (IC-1048 at column index 2)
+        // to avoid the unrelated looks_like_hierarchical_subrow heuristic.
+        let cells = vec![
+            vec![
+                "Order Date".into(),
+                "Suffix".into(),
+                "Item Code".into(),
+                "Description".into(),
+                "Status".into(),
+                "Unit Cost".into(),
+                "Freight".into(),
+                "Tax".into(),
+                "Total Billed".into(),
+            ],
+            vec![
+                "03/14/2024".into(),
+                "".into(),
+                "IC-1048".into(),
+                "".into(),
+                "Shipped".into(),
+                "".into(),
+                "".into(),
+                "".into(),
+                "".into(),
+            ],
+            vec![
+                "".into(),
+                "".into(),
+                "IC-1048".into(),
+                "WIDGET ASSEMBLY".into(),
+                "Shipped".into(),
+                "----".into(),
+                "...".into(),
+                "$0.00".into(),
+                "$482,110.40".into(),
+            ],
+        ];
+        let (cleaned, _) = clean_table_cells(&cells);
+        // Header + group-header row + 1 distinct detail row, none merged.
+        assert_eq!(cleaned.len(), 3);
+        assert_eq!(cleaned[2][5], "----");
+        assert_eq!(cleaned[2][6], "...");
     }
 
     #[test]
--- a/src/tables/format.rs
+++ b/src/tables/format.rs
@@ -307,9 +307,14 @@ fn clean_table_cells(cells: &[Vec<String>]) -> (Vec<Vec<String>>, Vec<String>) {
         let numeric_cells = non_first_cells
             .iter()
             .filter(|c| {
-                c.chars().all(|ch| {
-                    ch.is_ascii_digit() || ch == '.' || ch == '-' || ch == ',' || ch == ' '
-                })
+                // Strip a leading currency symbol so amounts like "$482,110.40"
+                // or "€1,200" are recognized as numeric data, not overflow text.
+                let c = c.trim_start_matches(['$', '€', '£', '¥']);
+                !c.is_empty()
+                    && c.chars().all(|ch| {
+                        ch.is_ascii_digit() || ch == '.' || ch == '-' || ch == ',' || ch == ' '
+                    })
+                    && c.chars().any(|ch| ch.is_ascii_digit())
             })
             .count();
         let looks_like_data_row = non_first_cells.len() >= 2
@@ -593,6 +598,69 @@ mod tests {
 
         assert_eq!(cleaned.len(), 4);
         assert_eq!(cleaned[2][0], "Mechanical");
+    }
+
+    #[test]
+    fn test_clean_table_cells_currency_data_row_not_merged() {
+        // A sparse/grouped table: the group-header row carries the order
+        // date/item code in column 0, and each detail row leaves column 0
+        // blank (row-spanning). Detail rows with dollar amounts must be
+        // recognized as real data rows and NOT merged as continuation text
+        // — the '$' prefix on numeric cells must not defeat the numeric
+        // check (see #229).
+        let cells = vec![
+            vec![
+                "Order Date".into(),
+                "Item Code".into(),
+                "Description".into(),
+                "Status".into(),
+                "Unit Cost".into(),
+                "Freight".into(),
+                "Tax".into(),
+            ],
+            vec![
+                "03/14/2024".into(),
+                "IC-1048".into(),
+                "".into(),
+                "Shipped".into(),
+                "".into(),
+                "".into(),
+                "".into(),
+            ],
+            vec![
+                "".into(),
+                "IC-1048".into(),
+                "WIDGET ASSEMBLY".into(),
+                "Shipped".into(),
+                "$482,110.40".into(),
+                "$0.00".into(),
+                "$9,215.75".into(),
+            ],
+            vec![
+                "".into(),
+                "IC-1048".into(),
+                "BRACKET SET".into(),
+                "Shipped".into(),
+                "$0.00".into(),
+                "$0.00".into(),
+                "$0.00".into(),
+            ],
+            vec![
+                "".into(),
+                "IC-1048".into(),
+                "CONTROL MODULE".into(),
+                "Shipped".into(),
+                "$31,905.22".into(),
+                "$4,100.00".into(),
+                "$0.00".into(),
+            ],
+        ];
+        let (cleaned, _) = clean_table_cells(&cells);
+        // Header + group-header row + 3 distinct detail rows, none merged.
+        assert_eq!(cleaned.len(), 5);
+        assert_eq!(cleaned[2][4], "$482,110.40");
+        assert_eq!(cleaned[3][4], "$0.00");
+        assert_eq!(cleaned[4][4], "$31,905.22");
     }
 
     #[test]
--- a/src/tables/format.rs
+++ b/src/tables/format.rs
@@ -604,30 +604,40 @@ mod tests {
     fn test_clean_table_cells_currency_data_row_not_merged() {
         // A sparse/grouped table: the group-header row carries the order
         // date/item code in column 0, and each detail row leaves column 0
-        // blank (row-spanning). Detail rows with dollar amounts must be
-        // recognized as real data rows and NOT merged as continuation text
-        // — the '$' prefix on numeric cells must not defeat the numeric
-        // check (see #229).
+        // *and* column 1 (Suffix) blank (row-spanning) — matching the real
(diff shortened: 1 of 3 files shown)


MADENIYOU:
Confirmed and fixed.

You're right — with \`IC-1048\` at column index 1, the row independently satisfies \`looks_like_hierarchical_subrow\`'s \`first_non_empty_col == Some(1)\` condition (a heuristic for hierarchical tables, unrelated to the numeric/currency check this PR changes), which alone keeps the rows apart regardless of whether the currency fix is present.

Fixed by restoring the fixture's actual shape from the real #229 repro: a blank \`Suffix\` column pushes \`IC-1048\` to column index 2 (defeating the \`== Some(1)\` condition), and the \`Total Billed\` column restores the numeric-cell majority (4 of 7 non-first cells) the currency fix is actually responsible for.

Verified by temporarily neutralizing the currency-stripping logic: the test now fails as expected (\`cleaned.len()\` collapses from 5 to 2, matching the original bug), confirming it exercises the intended path. Full suite: 869 tests passing, clippy clean on the changed file.

abimaelmartell:
Thanks for the focused fix. There is one behavioral regression in the current revision: requiring every numeric-shaped cell to contain a digit changes the existing treatment of placeholder cells such as `----` and `...`. Rows containing those placeholders can then be misclassified as continuation text and merged into a previous, distinct row.

Could you remove only the digit requirement and retain the non-empty guard after stripping currency symbols?

```rust
let c = c.trim_start_matches(['$', '€', '£', '¥']);
!c.is_empty()
    && c.chars().all(|ch| {
        ch.is_ascii_digit() || ch == '.' || ch == '-' || ch == ',' || ch == ' '
    })
```

`!c.is_empty()` still prevents a bare currency symbol from passing after it is stripped, while preserving the previous behavior for dash/ellipsis placeholders. Please also add a regression test covering rows with those placeholder cells.

src/tables/format.rs
```diff
@@ -307,9 +307,14 @@ fn clean_table_cells(cells: &[Vec<String>]) -> (Vec<Vec<String>>, Vec<String>) {
...
-                    ch.is_ascii_digit() || ch == '.' || ch == '-' || ch == ',' || ch == ' '
-                })
+                // Strip a leading currency symbol so amounts like "$482,110.40"
+                // or "€1,200" are recognized as numeric data, not overflow text.
+                let c = c.trim_start_matches(['$', '€', '£', '¥']);
+                !c.is_empty()
+                    && c.chars().all(|ch| {
+                        ch.is_ascii_digit() || ch == '.' || ch == '-' || ch == ',' || ch == ' '
+                    })
+                    && c.chars().any(|ch| ch.is_ascii_digit())
```
abimaelmartell: This digit requirement changes the established behavior for placeholder cells such as `----` and `...`: they no longer count as numeric-shaped, so a legitimate data row can be misclassified as continuation text and merged into the previous row. The `!c.is_empty()` guard above already rejects a bare currency symbol after stripping it, so this extra requirement should be removed.

```suggestion
```

MADENIYOU:
Confirmed and fixed, thanks for catching this.

You're right — the \`any(digit)\` requirement I added alongside the currency-symbol stripping narrowed the check beyond what it needs: \`----\` and \`...\` no longer counted as numeric-shaped, so rows using those placeholders could get misclassified as continuation text. Removed the digit requirement; \`!c.is_empty()\` after stripping the currency symbol already rejects a bare currency symbol on its own, matching pre-#229 behavior for placeholders.

Added \`test_clean_table_cells_dash_and_ellipsis_placeholders_still_count_as_numeric\`, using the same row layout as the currency test (\`IC-1048\` at column index 2, avoiding the unrelated hierarchical-subrow heuristic) with \`----\`/\`...\` in place of dollar amounts. Verified it actually catches the regression by temporarily reinstating the digit requirement — the test fails as expected (\`cleaned.len()\` drops from 3 to 2).

Full suite: 870 tests passing (1 new), clippy clean on the changed file.

PR #234, opened 2026-08-03
fix(extractor): detect two-column layout when the page also has a table

mohojojo:
## Problem

`detect_columns` (`src/extractor/layout.rs`) gates **both** its fallback column-detection strategies — relative-valley and XY-cut — behind `!page_has_table`:

```rust
if valleys.is_empty() && page_items.len() >= 100 && !page_has_table {
```

The intent (per the comment) is that a table's inter-column gaps can look like gutters in the projection histogram. But the side effect is that **any page which has a detected table _and_ genuine two-column prose loses column detection entirely** and collapses to single-column, Y-interleaved reading order.

This is common in real documents. A fund factsheet, for example, has left/right text columns sitting above a returns table. When the absolute-valley pass finds no empty gutter (justified text fills it), the relative-valley fallback is skipped because a table is present, so the two columns get read line-by-line interleaved:

```
**ÁLTALÁNOS INFORMÁCIÓK PIACI ÖSSZEFOGLALÓ:** Alapkezelő: Marketprog Asset
Management Zrt. Az alap HUF sorozatának árfolyama áprilisban 2,8%-kal emelkedett,
EUR sorozata Letétkezelő: Unicredit Bank Zrt. 3,6%-kal. Vezető forgalmazó: ...
```

(left-column labels shredded into right-column sentences).

## Root cause, confirmed

With `RUST_LOG=debug`, the same page yields opposite results depending on the `page_has_table` argument:

```
page 1: 0 valleys found but none passed validation      # page_has_table=true  → 1 column
page 1: 2 columns detected (boundaries: [294.4543])      # page_has_table=false → correct split
columns_have_prose: col [3..294]   ratio=0.87 avg_items=2.9
columns_have_prose: col [294..591] ratio=0.84 avg_items=1.9
page 1: relative valley detection found 2 columns
```

The relative-valley path finds the correct split at x=294 and **both sides pass `columns_have_prose`** — it's only the `page_has_table` gate that suppresses it.

## Fix

The relative-valley path already has its own table defense: `columns_have_prose` rejects splits whose sides look like tables/forms (`avg_items > 3.5` per line, low prose fill). That guard — not the page-level `page_has_table` flag — is the right protection, so the relative-valley fallback now runs regardless of tables.

The XY-cut fallback has **no** such prose guard, so it stays gated on `!page_has_table` to avoid reading a table's widest column gap as a page gutter.

## Verification

- **New regression test** `relative_valley_detects_columns_even_with_table_on_page`: two-column justified prose on a page flagged `page_has_table = true` must still split into two columns. Fails before this change (returns 1 column), passes after.
- Full suite green: `cargo test` (all existing tests unchanged), `cargo fmt --all -- --check`, `cargo clippy -- -D warnings`.
- Ran an unrelated 461-document corpus through `pdf2md` before/after: **only 4 outputs changed, all corrections** to the interleaving bug; the other 457 are byte-identical (the `columns_have_prose` guard keeps the change scoped to genuine two-column pages).

---
<a href="https://app.blacksmith.sh/firecrawl/codesmith/pdf-inspector/pr/234"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788364923&installation_model_id=11126&pr_number=234&repository=firecrawl%2Fpdf-inspector&return_to=https%3A%2F%2Fgithub.com%2Ffirecrawl%2Fpdf-inspector%2Fpull%2F234&signature=139946f0f030bd0a6c8a0a2c6891f556880ee81587e2e3b006dad97cda66a3a7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Fixes two-column detection on pages that also contain tables and stops justified prose from being misidentified as wide text tables. Reading order is corrected and phantom tables are avoided.

- **Bug Fixes**
  - Run relative‑valley fallback in `detect_columns` even when `page_has_table` is true; rely on `columns_have_prose` to reject table‑like splits, and keep XY‑cut gated behind `!page_has_table`.
  - Strengthen `is_paragraph_content`: detect near‑full rows of lowercase, multi‑word fragments as prose to avoid treating justified paragraphs as N‑column tables; real numeric/label tables remain unaffected.
  - Add/refresh regression tests: unit test for the prose heuristic and a snapshot (`marketprog-prestige-two-column`) that now reads columns correctly and removes the phantom table.

<sup>Written for commit 0ac0edb935f683e658c42031cfef2af54be6785b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/pdf-inspector/pull/234?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


mohojojo:
Added a real-document regression fixture: `tests/fixtures/marketprog-prestige-two-column.pdf` (a two-column fund factsheet with a returns table on the same page — the exact layout this fixes) plus its snapshot `tests/snapshots/marketprog-prestige-two-column.md` and `test_snapshot_marketprog_prestige_two_column`. The snapshot encodes the corrected reading order (left column ÁLTALÁNOS INFORMÁCIÓK → BEFEKTETÉSI POLITIKA contiguous, preceding the right column PIACI ÖSSZEFOGLALÓ) and fails if the interleaving regresses. Verified it fails on `main` and passes with the fix.

mohojojo:
Good catch — fixed in 0ac0edb rather than blessing the bad output.

**Root cause:** `has_table_like_content` bypasses the numeric-content check for any 3+ column table (to allow legitimate text-only tables like category lists). Justified paragraph text abuses this: the disclaimer's inter-word gaps align into vertical "columns", so a two-line paragraph is detected as an 8-column table — splitting one sentence into cells and re-emitting `Tájékoztatója…`, exactly the duplication/reordering you flagged.

**Fix:** `is_paragraph_content` now also recognizes this shape via a prose-row signal — a near-full row whose cells are mostly lowercase mid-sentence fragments with at least one multi-word cell. Two+ such rows ⇒ paragraph, so it's rejected as a table. Numeric or capitalized-label rows never match, so real tables are untouched (the stats table on the very same page still renders).

The snapshot is regenerated: the disclaimer now flows as a single paragraph, no phantom table, no duplication. Added unit tests `justified_prose_split_across_columns_is_paragraph` and `numeric_stats_table_is_not_paragraph`.

Verified on a 461-document corpus: removes justified-prose phantom tables from 52 documents and destroys no real data table (spot-checked the biggest deltas — every one was market-commentary or a garbled prose+data merge being correctly de-tabled, with the underlying data preserved as text).

PR #236, opened 2026-08-03
fix(tables): detect image-backed charts so legends stop becoming headings

mohojojo:
> **Stacked on #234.** This branch also contains the two commits from #234 (column detection). Until #234 merges, the diff below shows both; the net-new change here is the single commit `fix(tables): detect image-backed charts…`. Reviewing after #234 lands (or commit-by-commit) keeps it clean.

## Problem

Line/area charts are frequently emitted as a single **image XObject** with no bar geometry, so `detect_chart_regions` — which clusters filled `re` rects — never sees them. The chart's legend text (a line-swatch plus a series name, e.g. `----- Benchmark`) then flows as ordinary body text, and because it is short and bold it gets promoted to a heading, corrupting the outline:

```
## ----- Benchmark
## ----- Prestige Abszolút Hozamú Származtatott Részalap EUR sorozat
```

On a fund-factsheet corpus this produced 23 bogus `## -----` headings across 12 documents.

## Root cause

Confirmed with `RUST_LOG=debug`: the NAV chart is a raster image at `x=3..285, y=70..183`; the legend is live text at `y≈49..59` just below it (body-size **bold**, so it trips the isolated-bold / repeated-sidebar heading heuristics). No chart region is detected (`chart-aware=false`), because chart detection only knows bar charts.

## Fix

- **`detect_image_chart_regions`** — recognizes a chart-sized image as a chart region, **only when corroborated by an adjacent legend line-swatch**, so plain logos/photos are never mistaken for charts. The region is grown to cover the legend, so the existing `is_chart_adjacent_label` fencing (table + column masking, heading-sequence exclusion) applies to it.
- **Heading heuristic guard** — suppress heading promotion of a legend line (`starts_with_legend_swatch`) that sits inside a chart region. Deliberately **narrow**: real titles never begin with a swatch, so an over-large *bar*-chart region that happens to overlap a real heading does **not** demote it. (An earlier broad "no headings inside any chart region" guard demoted genuine titles on unrelated docs — hence the swatch gate.)

## Verification

- New unit tests: `legend_swatch_recognized`, `image_with_legend_swatch_is_a_chart_region`, `plain_image_without_legend_is_not_a_chart`, `small_image_with_swatch_is_ignored`.
- The `marketprog-prestige-two-column` snapshot (added in #234) now also captures the corrected output (no `## -----`).
- `cargo test` (721 + 142 + 2), `cargo fmt --all -- --check`, `cargo clippy -- -D warnings` all green.
- **Corpus (461 docs), before/after:** chart-legend headings **23 → 0**; **only 31 documents change** (all in the affected family); **zero prose text dropped** (the two "lost" tokens are glued words being correctly split); every other document byte-identical.

---
<a href="https://app.blacksmith.sh/firecrawl/codesmith/pdf-inspector/pr/236"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788370141&installation_model_id=11126&pr_number=236&repository=firecrawl%2Fpdf-inspector&return_to=https%3A%2F%2Fgithub.com%2Ffirecrawl%2Fpdf-inspector%2Fpull%2F236&signature=36462f0b14ac6fe7963a812d5dc7b04e49621f10e2a0dbade27ff6cf50b6ce4b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Detects image-backed charts and fences their legends so lines like "----- Benchmark" no longer become headings. Also restores two‑column reading order on pages with a table and stops justified prose from being misdetected as wide tables.

- **Bug Fixes**
  - Charts: detect chart-sized images only when an adjacent legend swatch is present and followed by a series name; grow the chart region to include the legend; suppress heading promotion for legend lines inside chart regions.
  - Columns: run the relative‑valley column split even if the page has a table; keep XY‑cut gated behind `!page_has_table` to avoid treating table gaps as gutters.
  - Tables: treat near-full rows of lowercase mid‑sentence fragments as prose, not tables; leaves numeric/label tables unchanged.
  - Tests: add legend‑swatch unit tests (including rejecting bare rules) and a two‑column factsheet snapshot; regenerate the snapshot to remove a phantom wide‑table; corpus impact — legend headings 23→0, justified‑prose phantom tables removed in 52 docs, no real tables lost.

<sup>Written for commit 98a3cc9952cedf686f1ad9c3fa67c707901bea2d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/pdf-inspector/pull/236?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


mohojojo:
Thanks — addressed in 98a3cc9.

**1 (P2, bare dash rule):** Valid. `starts_with_legend_swatch` now requires non-whitespace text after the dash run, so a decorative `-----` separator no longer corroborates a chart while `----- Benchmark` still does. Added `-----` / `--------` cases to the unit test.

**3 (P3, stale doc):** Fixed — the doc comment now states the region is the image bbox grown to enclose the matched legend.

**2 (P2, adjacent labels joining a table):** Real edge case, but I'd hold off on padding the claim region — here's the evidence. Across the 461-doc corpus, the chart change **removes** phantom table rows in 13 documents and **adds** them in exactly **1** (`MARKETPROG_Bond…Series_D`). On that one doc it's actually a net improvement: the performance returns that were previously mashed into a prose line now form a real 3-column table (`Időtáv | Befektetési jegy | Benchmark` + year rows), with one spurious 4th column of bleed (`KOCKÁZATI MUTATÓK…`).

Two reasons I don't think region-padding is the right fix:
1. The bleeding text (`KOCKÁZATI MUTATÓK AZ ELMÚLT 12 HÓNAPRA:`) is a **risk-section header, not chart furniture** — padding the chart region to swallow it would wrongly fence a legitimate heading, and if the pad is small enough to avoid that it won't reach this text anyway.
2. An earlier version of this PR broadly suppressed content inside chart regions and it **demoted real headings** on unrelated docs (over-large bar-chart regions overlapping titles) — I narrowed it specifically to avoid that. Growing the claim region reintroduces the same over-fencing risk.

Given it's 1/461 and net-positive there, I'd prefer to keep this PR scoped and treat frameless-chart axis-label fencing as a separate follow-up if it shows up as a real problem. Happy to revisit if you'd rather I add a conservative pad and re-run the corpus.

PR #268, opened 2026-08-05
fix(layout): enable sequential reading order for 2-column prose pages with <15 lines (#219)

### LINKED ISSUE #219: Two-column pages: columns are detected but markdown is emitted in raster order (lines interleaved)
## Summary

On newspaper-style two-column pages, columns are **detected** (`pages_with_columns` is populated) but the markdown is still emitted in **raster order**: each output line is the left-column line concatenated with the right-column line. The README documents "Automatic detection of newspaper-style columns, sequential reading order" — detection works, the sequential reordering does not appear to be applied.

On some real-world two-column pages the same layout is instead classified as a table (`pages_with_tables`), and the cells come out with the columns interleaved. Both symptoms destroy reading order.

- **Version:** `pdf-inspector` 0.2.6 (Python binding), Python 3.12.5, macOS 26.5 arm64
- **PDF type:** `text_based`, no OCR needed

## Reproduction

No sample file needed — this script builds the PDF (`pip install fpdf2`):

```python
from fpdf import FPDF

COL_W, COL_X = 85, (15, 110)
BODY = ("Paciente do sexo masculino, 62 anos, comparece ao servico com quadro de "
        "dispneia progressiva ha tres semanas, associada a edema de membros "
        "inferiores e ortopneia. Ao exame, estertores bibasais e turgencia jugular.")
ALTS = ["A) Iniciar diuretico de alca por via intravenosa.",
        "B) Solicitar ecocardiograma transtoracico.",
        "C) Prescrever betabloqueador em dose plena.",
        "D) Indicar cateterismo cardiaco de urgencia."]

pdf = FPDF(format="A4")
pdf.set_auto_page_break(False)

n = 1
for _ in range(2):                      # 2 pages
    pdf.add_page()
    for x in COL_X:                     # 2 columns per page
        pdf.set_xy(x, 20)
        pdf.set_font("helvetica", "B", 10)
        pdf.multi_cell(COL_W, 5, f"{n} - ", align="L")
        pdf.set_x(x)
        pdf.set_font("helvetica", "", 10)
        pdf.multi_cell(COL_W, 5, BODY + " A conduta mais adequada e:", align="J")
        for a in ALTS:
            pdf.set_x(x)
            pdf.multi_cell(COL_W, 5, a, align="L")
        n += 1

pdf.output("two-column-exam.pdf")
```

```python
import pdf_inspector
r = pdf_inspector.extract_pages_markdown("two-column-exam.pdf")
print(r.pages_with_columns)   # [1, 2]
print(r.pages[0].markdown)
```

## Actual output (page 1)

```
## 1 -2 -

Paciente do sexo masculino, 62 anos, comparece ao Paciente do sexo masculino, 62 anos, comparece ao servico com quadro de dispneia progressiva ha tres servico com quadro de dispneia progressiva ha tres semanas, associada a edema de membros inferiores semanas, associada a edema de membros inferiores e ortopneia. Ao exame, estertores bibasais e e ortopneia. Ao exame, estertores bibasais e turgencia jugular. A conduta mais adequada e: turgencia jugular. A conduta mais adequada e:

A) Iniciar diuretico de alca por via intravenosa. A) Iniciar diuretico de alca por via intravenosa.
B) Solicitar ecocardiograma transtoracico. B) Solicitar ecocardiograma transtoracico.
C) Prescrever betabloqueador em dose plena. C) Prescrever betabloqueador em dose plena.
D) Indicar cateterismo cardiaco de urgencia. D) Indicar cateterismo cardiaco de urgencia.
```

Both items collapse into `## 1 -2 -`, and every body line is the left line + the right line glued together.

## Expected output

Column 1 fully, then column 2:

```
## 1 -

Paciente do sexo masculino, 62 anos, ... A conduta mais adequada e:

A) Iniciar diuretico de alca por via intravenosa.
...

## 2 -

Paciente do sexo masculino, 62 anos, ... A conduta mais adequada e:

A) Iniciar diuretico de alca por via intravenosa.
...
```

`pdftotext -layout` (poppler) keeps the two columns separable on the same files.

## Real-world impact

Measured on a corpus of 26 two-column Brazilian medical residency exam PDFs (text-based, no OCR). A regex parser that splits questions on the `N -` marker at line start recovers all items from `pdftotext -layout`, but from `extract_pages_markdown` the split dies at the first affected page:

| PDF | items expected | items recovered |
|---|---|---|
| SUS-SP R+ 2021 | 50 | 24 |
| SUS-SP R+ 2024 | 50 | 24 |
| UNICAMP R+ 2022 | 70 | 27 |
| UNIFESP R+ 2026 | 50 | 16 |
| USP-SP R+ 2026 | 120 | 96 |

The other 21 files (single-column, or two-column pages that happened not to trigger it) parse identically to poppler once markdown syntax is normalized, and the answer-key grids actually come out **better** as markdown tables. So this is the one blocker for using pdf-inspector as the extractor in that pipeline.

In the SUS-SP 2021 case the offending page went down the table path instead, producing rows like:

```
|24 - 25 -|respectivamente:|B) Indicar toracoscopia diagnostica. C) Indicar drenagem toracica. ...
```

i.e. two item markers in one cell, and alternatives from different items merged into a third — same root cause, different renderer.

## Notes

Happy to test a patch against the 26-file corpus. A public escape hatch would also help even before the ordering is fixed — the Python binding currently exposes only `pages`, so there is no way to opt out of table detection or to request a layout-preserving text mode. The Rust `PdfOptions` / `ProcessMode` mentioned in the README are not reachable from Python 0.2.6.

AyushDubey23:
### Summary
Fixes #219 where two-column pages with fewer than 15 lines per column (such as 2-column medical residency exam pages, short articles, or final chapter pages) were rejected by `is_newspaper_layout` and emitted in raster/Y-interleaved order, concatenating left and right column text onto single lines.

### Cause
`is_newspaper_layout` used an unconditional check `if min_lines < 15 { return false; }`. For 2-column pages with 5 to 14 lines per column:
1. `layout.rs` treated the page layout as tabular, Y-interleaving lines across columns and gluing left-column lines to right-column lines (e.g. `## 1 -2 -`).
2. In table detection (`try_build_table_from_columns`), `is_newspaper_layout` returning `false` caused 2-column prose to occasionally fall down the table path and produce merged table cells.

The `min_lines < 15` guard was originally added to prevent 4+ column borderless data tables (e.g. 5-column technical spec tables with 5 lines) from being treated as multi-column prose.

### Changes
- Scoped `min_lines < 15` strictly to 4+ column layouts (`if columns.len() >= 4 && min_lines < 15 { return false; }`).
- 2-column and 3-column prose pages with 5–14 lines per column and balanced line counts (`balance_ratio > 0.7`) are now correctly recognized as newspaper layout and emitted sequentially (Column 1 top-to-bottom, then Column 2).
- Added `two_column_exam_layout_retained_sequentially` unit test in `src/extractor/layout.rs` and updated layout classification tests.

### Verification
- `cargo fmt`
- `cargo test --lib` (799 tests passing)
- `cargo build --release`

---
<a href="https://app.blacksmith.sh/firecrawl/codesmith/pdf-inspector/pr/268"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788519756&installation_model_id=11126&pr_number=268&repository=firecrawl%2Fpdf-inspector&return_to=https%3A%2F%2Fgithub.com%2Ffirecrawl%2Fpdf-inspector%2Fpull%2F268&signature=0e7533c2ea3d21c8d71a42fd9067315e07449ffbf3347610db21556078e54f66"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Fixes #219 by classifying short, balanced 2–3 column prose pages (<15 lines/col) as newspaper so text is read column-by-column, and rejecting table hypotheses for these pages to avoid accidental Markdown tables. Sparse 4+ columns and short unbalanced columns remain tabular.

- **Bug Fixes**
  - Relaxed the `<15 lines` guard only for balanced 2–3 column layouts (`balance_ratio > 0.7`); short unbalanced or 4+ columns are not newspaper.
  - Sidebar detection runs first to prevent false negatives; detects narrow sparse sidebars vs body.
  - Markdown: re-check 2–3 column `TableKind::Data` candidates and reject “newspaper” table hypotheses to prevent table output.

- **Tests**
  - Added balanced 3-column and short unbalanced 2-column cases; kept the 2-column exam sequential test.
  - Updated tabular detection to a sparse 4-column case; refreshed snapshots for sequential reading and table rejection.

<sup>Written for commit 599508512098f5f55a34494fb23d144d8640704d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/pdf-inspector/pull/268?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rmacy:
I tested the exact PR head, `4a071a3`, against a separate synthetic short-column case. The prose-classifier change helps the reported exam layout, but this case still falls through to the false-table path from #219.

Fixture shape:

- one full-width title
- two column headers
- eight aligned rows per column
- independent sequences: `LEFT-01` through `LEFT-08`, then `RIGHT-01` through `RIGHT-08`

Expected reading order:

```text
PRIMARY ACTIVITY
LEFT-01 ... LEFT-08
ACCOUNT MESSAGES
RIGHT-01 ... RIGHT-08
```

Actual output at this PR head:

```markdown
## PRIMARY ACTIVITY ACCOUNT MESSAGES

|LEFT-01 $11.00|RIGHT-01 2021|
|---|---|
...
```

The CLI reports `pages_with_tables: [1]`. Parser-neutral anchor scoring remains 57.89% because the body-font heuristic claims a two-column table before sequential reading order applies.

A local fallback that rejects this specific table hypothesis and reads each indexed sequence down scored 100% on the fixture. I would add this shape as a regression before merge. The current classifier fix and the false-table fix are complementary.

The fixture is generated and contains no external document content.

AyushDubey23:
Thanks for the detailed review and for sharing the synthetic fixture!

We have pushed commit `5995085` which resolves all review items and the false-table claim:

1. **P2 (Unbalanced short layouts)**: Scoped the short-layout (<15 lines) relaxation in `is_newspaper_layout` strictly to balanced column line counts (`balance_ratio > 0.7`). Short unbalanced layouts (e.g. 5 vs 10 lines) continue to be rejected as non-newspaper.
2. **P3 (3-column test coverage)**: Added unit test `three_column_balanced_sparse_detected_as_newspaper` verifying that balanced 3-column pages with 5–14 lines per column return `true`.
3. **False-table claim on short 2-column prose**: Added `is_newspaper_prose_table` in `src/markdown/mod.rs` to evaluate candidate body-font tables against column layout classification. If the text items form a 2-column or 3-column newspaper prose layout, the false body-font table hypothesis is rejected so text is emitted in sequential reading order (`Column 1` top-to-bottom, then `Column 2`).

All 805 unit and integration tests pass cleanly, and the synthetic fixture now scores 100% with sequential reading order (`PRIMARY ACTIVITY` section followed by `ACCOUNT MESSAGES` section).

xNarkon:
I tried this PR on some Polish government gazettes (Dziennik Ustaw) that hit #219, hoping it would
fix them. It doesn't - but I went digging and found where they actually break, so I figured I'd
share rather than sit on it.

Quick check first: built this branch and `main`, ran both over the two PDFs, and compared each page
against `pdftotext` as a reading-order reference. Output is identical between the two builds. The
bad pages stay bad - 0.49 and 0.64 word-sequence similarity on the two-column pages, 0.61 on one
page that gets mistaken for a table. That makes sense once you look at the guard being relaxed
here: it's scoped to pages with under 15 lines per column, and these pages have 48–61. They never
get that far.

What happens to them instead, on page 4 of Dz.U. 2010 poz. 931:

`merge_text_items` glues the two columns together. The left column's line ends at x=298.9, the
right column starts at 303.6, so the gap is 4.7pt - and `x_gap_max` is `font_size * 0.5` = 4.75pt.
It just squeaks under. The result is a sentence that exists in neither column:

    …dla systemów i urządzeń przeznaczonych do powinien zapewnić, aby osoby pracujące zostały
    użytku w przestrzeniach zagrożonych wybuchem.

Justified gazette columns run right up to the gutter, so the gap between columns ends up the same
size as a wide word space. There's a second variant of this where the run arrives already spanning
both columns, because the `TJ` offset sits under `column_gap_threshold`.

Then `detect_columns` can't save it either: `MIN_GUTTER_WIDTH` is 8.0 and these gutters are 4.7–7pt.
The histogram does find the valley (246–252pt on that page), it just gets dropped on width, so the
page reads as a single column and the lines come out interleaved.

Here's the annoying part. I patched all of that locally - verified no item crosses the gutter,
`detect_columns` returns 2 columns on all four pages, `is_newspaper_layout` returns true - and
`extractPagesMarkdown` output didn't change by a single byte. So whatever decides the order for
these pages lives further down in `src/markdown/mod.rs`, not in the newspaper branch in `layout.rs`.
That's where I ran out of road.

One warning for whoever picks this up: my patch broke 2 of the 162 snapshot tests
(`nexo-price-en` and `real-estate-pricing`), both table documents. Not surprising in hindsight - a
table's column gap looks exactly like a prose gutter if all you measure is horizontal emptiness. So
a gutter-based fix probably needs the table signals you already have, or a check that the gap runs
the full height of the body (tables get interrupted by spanning header rows, real gutters don't).

Attaching both PDFs. Page 7 of the second one is a good control - its voltage table extracts
correctly today, so it shouldn't regress.

Happy to open a separate issue if you'd rather keep this PR on the short-page case.

[ BHP przy urządzeniach energetycznych.pdf](https://github.com/user-attachments/files/31046262/BHP.przy.urzadzeniach.energetycznych.pdf)
[Atmosfera wybuchowa.pdf](https://github.com/user-attachments/files/31046263/Atmosfera.wybuchowa.pdf)

AyushDubey23:
@xNarkon Really appreciate the detailed breakdown and the Polish gazette fixtures. This makes the failure modes super clear.

The main issues are:

1. **`x_gap_max`**: On justified 2 column pages, the ~4.7pt gutter falls just under the ~4.75pt merge threshold, so left and right text gets merged.

2. **`MIN_GUTTER_WIDTH`**: The 8pt minimum misses valid 4.7 to 7pt gutters, causing `detect_columns` to treat the page as single column.

3. **Downstream partitioning**: `split_side_by_side` misses these narrow gutters, so `to_markdown` falls back to Y sorting even when `is_newspaper_layout` is correct.

I agree #219 should stay focused on the <15 line classification fix.

We’ll open a follow up issue for **dense justified 2 column pages with narrow gutters (<8pt)** covering gutter aware merging, full span gutter validation to avoid table regressions, and column aware grouping to preserve the correct reading order.

Thanks again for digging into this. This gives us a pretty clear roadmap for the follow up fix! 🙌

PR #317, opened 2026-08-09
fix(markdown): collapse exact duplicate text overlays

rmacy:
## Summary

- collapse exact duplicate text and form-field paint operations before Markdown conversion
- keep text at different positions, fonts, sizes, or semantic styles
- preserve every source operation in `extract_text_with_positions`
- keep the precomputed page-number mask aligned after the filter

## Regression proof

`exact_duplicate_overlay_is_emitted_once_in_markdown` failed on `436af97` before the fix.

The synthetic PDF paints `Purchase $19.95` twice at one position and once at another position. The unfixed Markdown contained three copies. The expected semantic Markdown contains two.

After the fix:

- Markdown contains two copies, one for each position
- positioned extraction contains all three source operations

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo clippy --offline -- -D warnings`
- [x] `cargo test --offline` — 992 passed
- [x] `cargo build --release --offline`
- [x] manual generated duplicate-overlay fixture through `pdf2md`

The test fixture is generated and contains no external document content.

### DIFF
--- a/src/markdown/mod.rs
+++ b/src/markdown/mod.rs
@@ -1133,6 +1133,61 @@ pub(crate) fn to_markdown_from_items_with_rects_and_lines(
         return String::new();
     }
 
+    // Keep positioned extraction faithful to the source operations. Semantic
+    // Markdown needs only one copy of the same text painted at identical
+    // coordinates.
+    #[derive(Eq, Hash, PartialEq)]
+    struct OverlayKey {
+        page: u32,
+        is_form_field: bool,
+        text: String,
+        x: u32,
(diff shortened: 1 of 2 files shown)


rmacy:
Fixed in c5b5e2e. OverlayKey now includes mcid, so exact duplicate paints only collapse within the same tagged node. The regression test covers identical text and geometry with distinct MCIDs. Validation passed: cargo fmt --check, cargo clippy -- -D warnings, cargo test, and cargo build --release.

PR #328, opened 2026-08-09
Bound column-detection histogram and harden coordinate handling

abimaelmartell:
## Summary

`detect_columns()` in `src/extractor/layout.rs` derives its page bounds from text-item coordinates, which come straight from the content-stream text matrix (`Tm`/`Td`). A document can therefore place a run at any position, and those bounds fed two problems:

1. **Unbounded allocation.** The histogram bin count (`page_width / BIN_WIDTH`) had no upper bound, so an extreme coordinate drove an arbitrarily large `vec![0u32; num_bins]`.
2. **Poisoned thresholds.** Gutter margins, the spanning-item width cut-off and the XY-cut margin are all fractions of `page_width`, so one far item shrank the effective detection window and silently disabled column detection for the whole page.

## Changes

- Clamp the bin count with a `MAX_BINS = 65_536` ceiling (~128k points at `BIN_WIDTH = 2.0`). Kept as a hard allocation bound that does not depend on the heuristics below staying correct.
- Exclude items at non-finite coordinates when folding the bounds. A NaN is otherwise silently dropped by `f32::min`/`f32::max` while an inf propagates through and can escape as a `ColumnRegion` boundary. Bad items are skipped individually rather than failing the page.
- Trim far outliers when the span exceeds `MAX_PAGE_EXTENT = 14_400` units, re-deriving the bounds from items clustered around the median. **Both item edges are checked**, since a malformed *width* at an ordinary position poisons `x_max` exactly as a malformed position poisons `x_min`.
- Only trim when the far items are a small minority (`MAX_TRIM_FRACTION = 10%`). A genuinely large-format page has content spread across its full width, so it keeps its true bounds rather than being reduced to the median cluster.
- Return no columns when every item sits at a non-finite coordinate (callers already treat `len() <= 1` as single-column).
- Use `saturating_sub` for the sibling `Vec::with_capacity` in `split_column_stragglers()` as cheap defence-in-depth (capacity is only a hint).

Outliers never lose their text: column assignment buckets by greatest overlap and falls back to column 0, so a zero-overlap item is still placed.

### On `MAX_PAGE_EXTENT`

14_400 units is **not** a format limit, and the comment now says so. PDF 2.0 sets no page-size cap, and since PDF 1.6 `UserUnit` scales a page's physical size independently of its coordinates. 14_400 is the traditional Acrobat architectural limit (PDF 1.7 Annex C), which makes it a reasonable "wider than any ordinary page" mark in coordinate space. It is documented as a heuristic, and the 10% minority rule is what actually protects valid large-format documents.

## Testing

- `cargo fmt`, `cargo clippy -- -D warnings`, `cargo test` all pass (852 lib + 152 integration + doc tests).
- New tests: unbounded-allocation guard; non-finite values never leaking into emitted bounds (`+inf`, `-inf`, NaN); all-non-finite page yields no columns; a two-column page keeping both columns across six malformed-item shapes (NaN/inf/far positions **and** huge/inf widths); a genuinely wide 40_000pt layout keeping its true bounds; a wide-but-ordinary 8_000pt page not trimmed.
- The wide-layout test was verified to be load-bearing: relaxing `MAX_TRIM_FRACTION` to 1.0 makes it fail (right edge 33_390 instead of ~40_000).
- Verified before/after on a crafted 1.2 KB PDF: the previous build aborted while trying to reserve terabytes; now it converts normally at low RSS, with NaN, ±inf and float-saturation variants all clean.

### Output-stability check

The `pdf-evals` snapshot suite is not reachable from this environment, so `bench.py test` / `bench.py score` have not been run. As a substitute, binaries built from `main` and from this branch were run over all 27 fixture PDFs in `tests/fixtures/`:

- **27/27 byte-identical output.**
- **0/27 reach the outlier-trim path** (checked via the new debug log).

That matches the design: every new branch is gated behind a condition well-formed PDFs cannot meet — `MAX_BINS` needs a span above 131_072pt, trimming needs one above 14_400pt plus a <10% stray minority, and the non-finite filters need a NaN/inf coordinate. Snapshots should only move for a document that genuinely has a few stray off-page items, where the expected change is recovering column detection that is silently lost today.

<div><a href="https://cursor.com/agents/bc-019fe7a2-665a-7879-929e-ffdba3d33fe5?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-019fe7a2-665a-7879-929e-ffdba3d33fe5&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

---
## Summary by cubic
Bound and scale the column-detection histogram, harden geometry handling, judge run widths relative to page content, and require geometric evidence before trimming detached items. This prevents huge allocations and bad bounds while preserving real layouts, including very wide pages and long runs.

- **Bug Fixes**
  - Cap histogram bins at 65,536 and derive `bin_width` from `page_width` so coverage spans the full page; gutters stay correct past ~131k points.
  - Skip items with non-finite geometry (NaN/±inf positions or left+width); all-non-finite pages yield no columns; region edges are always finite.
  - Replace median trim with gap-based clustering on positions; only drop clusters split by >1 page of empty space and ≤10% of items; judge widths relative to the page’s content spread (+1 page) so valid large pages and long runs keep true bounds; check both edges so huge widths can’t poison bounds.
  - Use saturating capacity in `split_column_stragglers()`.

<sup>Written for commit 83d378ebe8457040a5d9b0180c956750b29b3876. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/pdf-inspector/pull/328?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/extractor/layout.rs
+++ b/src/extractor/layout.rs
@@ -41,22 +41,35 @@ pub(crate) fn detect_columns(
     }
     debug!("page {}: detect_columns: {} items", page, page_items.len());
 
-    // Find page bounds
-    let x_min = page_items.iter().map(|i| i.x).fold(f32::INFINITY, f32::min);
+    // Find page bounds. Coordinates come straight from the content-stream text
+    // matrix, so a malformed document can place an item at a non-finite
+    // position. Such items are excluded from the bounds rather than failing the
+    // whole page: one bad glyph should not disable column detection, and the
+    // emitted regions must never carry a NaN/inf boundary to callers.
+    let x_min = page_items
+        .iter()
+        .map(|i| i.x)
+        .filter(|x| x.is_finite())
+        .fold(f32::INFINITY, f32::min);
     let x_max = page_items
         .iter()
         .map(|i| i.x + effective_width(i))
+        .filter(|x| x.is_finite())
         .fold(f32::NEG_INFINITY, f32::max);
 
-    // Cap the histogram size. Bounds are derived from text-item coordinates,
-    // which come straight from the content-stream text matrix and can be placed
-    // at any coordinate, so a crafted PDF can otherwise force an arbitrarily
-    // large `page_width` and an unbounded `vec![0u32; num_bins]` allocation.
-    // 65_536 bins covers ~128k points at BIN_WIDTH 2.0 — roughly 9x the largest
-    // legal page — so this never clamps a real layout. A non-finite width
-    // (NaN/inf coordinate propagating through the folds) short-circuits here.
+    // Every item sat at a non-finite coordinate, so there is no usable layout.
+    if !x_min.is_finite() || !x_max.is_finite() {
+        return vec![];
+    }
+
+    // Cap the histogram size. Because the bounds above are attacker-influenced,
+    // an unclamped `page_width / BIN_WIDTH` lets a crafted PDF force an
+    // arbitrarily large `vec![0u32; num_bins]` allocation. 65_536 bins covers
+    // ~128k points at BIN_WIDTH 2.0 — roughly 9x the largest legal page — so
+    // this never clamps a real layout.
     const MAX_BINS: usize = 65_536;
 
+    // Finite bounds can still subtract to infinity at the extremes of f32.
     let page_width = x_max - x_min;
     if !page_width.is_finite() || page_width < 200.0 {
         return vec![ColumnRegion { x_min, x_max }];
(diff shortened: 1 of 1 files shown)


abimaelmartell:
@cubic-dev-ai

abimaelmartell:
@cubic-dev-ai

PR #331, opened 2026-08-10
fix(rtl): logical order for RTL text, via UAX #9 and the structure tree

j0ck4:
Fixes #58.

Hebrew came out reversed word by word, and Arabic only worked when the producer
wrote presentation forms. Same cause for both: the reordering pass ran only when
Arabic presentation forms were present, and Hebrew never has them — it needs no
contextual shaping, so the reordering never ran for it at all.

## Why the order does not have to be guessed

`expand_ligatures` only ever sees the text of a single text-showing operator, and
within one such run the pen advances left to right by each glyph's width. A
multi-glyph RTL run therefore has to be stored in visual order for the page to
render correctly — there is no other way to lay it out with one advancing run.
So this is not a heuristic about the producer; it follows from how PDF places
text.

Confirmed on the operators of a LibreOffice file: the heading `ישראל ישראלי`
is a single `Tj` with twelve glyphs at one `Td` position.

The case this loses is a producer that writes logical order inside one run. Such
a file renders mirrored in every viewer, and matching it would mean breaking the
files that render correctly.

## The replaced test

`no_reversal_for_base_arabic` asserted that base Arabic passes through unchanged,
on the assumption that it arrives in logical order. Absence of presentation forms
says nothing about order — only that the producer wrote base codepoints, which
LibreOffice, mPDF and iTextSharp all do while laying out visually. The test
protected the bug, so it is replaced by one that states the rule.

## Second fix: line direction

`is_rtl_text` used a plain majority of letters, which let one embedded Latin
token flip a whole line. On a CV contact line — Hebrew role and city next to
`israel@example.com` — the address matched the Hebrew letter for letter, the tie
fell to LTR, and the line came out ordered backwards. Latin now has to outweigh
the RTL text twice over before a line counts as left-to-right.

## Measurements

Same binary before and after, counting Hebrew section words in forward and
reversed spelling:

| File | before (correct/reversed) | after |
|---|---|---|
| Hebrew CV (DevOps, LibreOffice) | 0 / 10 | **10 / 0** |
| Hebrew CV (second, different producer) | 0 / 9 | **9 / 0** |
| Synthetic Hebrew | 0 / 7 | **7 / 0** |
| Synthetic Arabic | 0 / 1 | **3 / 0** |
| English CV | 0 / 0 | 0 / 0 — untouched |
| CV that already extracted correctly | 7 / 2 | 7 / 2 — no regression |

All 1005 tests pass.

## Known and not addressed here

Neutral characters still land in the wrong place on mixed lines: `דוא"ל:` comes
out as `:ל"דוא`, and separators bunch at the start of a contact line. Three or
four places per document; words and word order are correct, and emails and phone
numbers still extract by regex.

I tried the fuller fix — moving reordering off individual items and onto the
assembled line, with neutral resolution — and measured it as **worse**: 9/2 and
9/1 against 10/0 and 9/0 above. It breaks where a line is split into runs with
their own styling; an underlined email address is its own item, and the run
boundaries defeat the neutral resolution. Doing it properly needs real bidi over
a sequence of items while preserving styling, which is a separate piece of work.

Also unrelated to ordering: `الخبرة` extracts as `الخربة` (two letters swapped).
That predates this change — it was present before, merely reversed as well. It
looks like a separate defect in the Arabic glyph mapping. Happy to open a
separate issue if useful.

---
## Summary by cubic
Restores logical order for RTL text using UAX #9 with document-level direction, fixing reversed Hebrew and base Arabic while keeping embedded Latin and digits readable. Also prevents merged columns by using tagged PDF reading order and, when untagged, inferring a local column band from aligned row splits. Fixes #58.

- **New Features**
  - Replace heuristics with UAX #9 reordering via `unicode-bidi`; paragraph level comes from the document’s dominant direction.
  - Tagged PDFs: rebuild lines by structure-tree MCID reading order (98% coverage gate) so columns don’t interleave.
  - Untagged pages: detect a local two‑column band by clustering aligned row splits (with prose checks on both sides) and use it when no image anchor exists.

- **Bug Fixes**
  - Apply document direction to item, line, and layout sorting; drop per-line letter counting.
  - In RTL lines, flip glyph‑per‑item Latin/digit runs back to LTR so emails and phones read forward.
  - Reversal is grapheme‑cluster aware; digits never count as RTL; `ActualText` is normalized but not reversed.

<sup>Written for commit f5a281e93129c3dae32f2f505822812d3556f1ae. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/pdf-inspector/pull/331?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


j0ck4:
Thanks — all six findings were real. Every one reproduced before it was fixed, and the tests now assert the corrected behaviour.

**A note on how the first attempt at those tests failed**, because it bears on how the rest should be read. I initially wrote them as round trips: take a logical string, run it through `reverse_visual_rtl`, feed the result back to `expand_ligatures`, assert the original comes back. Three of the four passed while the code was still broken — a round trip is symmetric, so the same wrong classification applies in both directions and cancels itself out. The tests now build the visual form independently, and all four failed immediately, exactly as described in the review.

Two of my replacement fixtures were also badly chosen: `שלום François Müller` is four Hebrew letters against fourteen Latin, which is structurally the same as your `Hello של world` — a case that must *not* be reordered. Counting cannot tell them apart. The fixtures now use genuinely RTL-dominant lines.

---

**1 & 2 (P1) — direction classification.** `is_ascii_alphanumeric` was doing work it cannot do. `François` and `Müller` are not ASCII, so they were classified as RTL and came out as `llerüM oisçFran`. Arabic-Indic digits live inside the Arabic block, so a phone number reversed. Letters and digits of any script are now classified by their own properties, and digits are never RTL — in `reverse_visual_rtl` and in `restore_ltr_runs` alike, which covers your second review's P1 about glyph-per-item Arabic-Indic numbers.

**3 (P2) — over-application to LTR-dominant runs.** Confirmed exactly as written: `Hello של world` became `world לש Hello`. The trigger now requires the Latin content to outweigh the RTL text twice over before a run counts as left-to-right — the same rule `is_rtl_text` already applies to a line, so the two agree instead of contradicting each other.

**4 (P2) — combining marks.** Reversal walks grapheme clusters instead of scalars, so vocalisation stays on its letter; `שָׁלוֹם` was losing its points to the neighbouring character.

**2.2 in the first review (P1) — `ActualText`.** You were right that the comment overstated where this helper is reached from. `ActualText` is a logical-order replacement by specification, so it now gets cleanup and NFKC through a separate entry point but no reordering. Tagged PDFs were being reversed against the producer's explicit intent.

**Second review, P2 — missing test.** Added: an RTL line holding a glyph-per-item Latin address, asserted through `merge_text_items` rather than `sort_line_items`, so it exercises `restore_ltr_runs` itself.

---

1010 tests pass. On real Hebrew CVs the name, the email address and the phone number are now correct at the same time — none of the three were before this branch. An English CV is byte-identical to the previous output, and a CV that already extracted correctly is unchanged.

Still not addressed, and I would rather say so than let it be discovered later: neutral characters on mixed lines land in the wrong place, so `דוא"ל:` comes out as `:ל"דוא` and separators bunch at the start of a contact line. I tried moving the reordering off individual items and onto the assembled line with neutral resolution, measured it, and it came out **worse** — 9/2 and 9/1 against 10/0 and 9/0 on the same corpus — because a line split into runs with their own styling defeats the neutral resolution at the run boundaries. Doing it properly needs real bidi over a sequence of items while preserving styling, which is a separate piece of work rather than a follow-up commit here.

j0ck4:
The approach in this PR has changed, so the title no longer describes it well and the earlier commits are best read as history rather than as the proposal.

**What forced the change.** After the last round I tested on a wider corpus — 404 real CVs — and hit a line the heuristic could not handle: `שירותי מחשוב System Administrator Genie 2013-2017`, where the Hebrew stayed reversed. Fixing it by moving the threshold would have re-broken `Hello של world` from your review. The two have nearly the same ratio of Hebrew to Latin and require opposite treatment, so no count taken inside a run can separate them. That is not a tuning problem; it says the input to the decision is wrong.

**So the heuristic is gone.** Direction now comes from the paragraph level, as UAX #9 prescribes. Rule P2/P3 derives that level from the first strong character in logical order — exactly what a visual-order stream does not have — but the standard permits a higher-level protocol to supply it (HL1). The protocol here is the document's own direction, measured across all extracted text. At document scale the signal is stable: a Hebrew CV counts 1210 strong RTL characters against 492 LTR, an English one 0 against 1436. At line scale it is noise, which is what the thresholds were failing to paper over.

Everything downstream is `unicode-bidi`: classes, weak types, neutral resolution, levels, L2. The four defects you found — non-ASCII Latin, Arabic-Indic digits, combining marks, over-application to LTR-dominant runs — are handled by the standard rather than by my code. One post-step is still mine: L2 reorders characters, so a base letter and its mark change places, and they are put back together afterwards.

It also turned out that character order inside an item was only the first of three places. A word is regularly split across items, so line ordering decides it; and headings are assembled a third time in the layout engine. All three now take direction from the page.

**Second commit, separate concern.** `structure_tree.rs` already walked the tree to find tables, and `TextItem` already carried its MCID; the two were never connected. The traversal now also numbers each marked-content span, and lines are rebuilt from that numbering, which is the reading order ISO 32000-1 §14.8.2.3 defines. Two columns at the same height stop merging, because their pieces are not neighbours in the tree however close they sit. Gated on near-complete coverage — mixing a declared order with guesses about the remainder yields a third order, wrong in its own way.

**Measurements**, same binary before and after, on 404 real CVs:

| | before | after |
|---|---|---|
| reversed words | 24 | **20** |
| files with reversed words | 11 | **7** |
| files with a contact wedged mid-sentence | 54 | **51** |

1014 tests pass. Of 358 PDFs, 249 are tagged; on a 150-file sample the structure-tree gate opens for 79.

**Not addressed, deliberately.** Untagged multi-column pages — 61 of that sample. The crate already has `extractor/reading_order.rs`, but it is gated on image evidence and stays silent on a plain two-column CV. Widening it along the lines of PyMuPDF's column detection is a separate piece of work, and I would rather propose it on its own than bury it here.

Happy to split this into two PRs if you would prefer the bidi change and the structure-tree change reviewed separately.

j0ck4:
One more commit, and a report on something I tried and threw away.

**`feat(order): infer a local column band from aligned row splits`** — the region graph could previously only be reached through image evidence, so a page whose columns are marked by nothing but a gutter fell back to the whole-page projection, which by construction cannot see a band occupying part of the page. `local_flow_below_full_width_image` already clustered aligned row splits to locate the band; this offers the same clustering without requiring the anchor, keeping both guards — four rows splitting at the same x, and both sides passing `side_is_prose`.

Three snapshots changed and are updated in that commit, because the previous output was the defect rather than the baseline:

- `real-estate-pricing` line 45 read `more about investing in tax losses than burst, cap rates spreads steadily com- r` — two column fragments spliced — and now reads `more about investing in tax losses than real estate cash streams.`
- `thermo-freon12` had `S.A., Lemmon, E.W., and Peskin, Vf = Fluid (liquid) specific volume`, an author list welded to a symbol definition from the neighbouring column; the two are now separate and the symbol list is emitted as its own block.

**What I tried next and reverted.** CV sidebars are still merged into the body text, and neither the projection nor the row-split band reaches them: instrumenting one such page showed the band inference runs, sees 52 rows, 40 of them with two or more items, and finds no qualifying split — the sidebar keeps its own line spacing, so its lines never share a row with the body. That is what `aligned_row_split` requires and what journals have and CV sidebars do not.

So I implemented block-based detection in the spirit of PyMuPDF's `multi_column.py`: cluster items into blocks by proximity on both axes rather than by rows, then treat blocks that overlap vertically and are disjoint horizontally as neighbouring columns. It works on the target page — the sidebar and the body separate cleanly. I still reverted it, for two reasons.

The first is a regression: letter-spaced display type is read as columns. `**C O M M E R C I A L R E A L E S T A T E**` came out as `**R E A L**`, because the tracking between letters looks exactly like a column gutter at that scale. That needs a real guard, not another threshold.

The second matters more. I could not demonstrate that the change helped, because the metric I had been steering by was wrong. I was counting files where a contact detail sits mid-sentence, treating that as the signature of merged columns. Inspecting the hits showed most of them are ordinary contact lines — `כפר סבא, 71/10תל חי | 24110074 | ת.ז.: …@gmail.com | 052-…` is a header, not a defect. On 392 CVs the count sat at 52 with the block pass and 52 without it, which told me nothing either way.

Rather than tune an algorithm against a broken measurement, I would rather build the measurement first — and the tagged files make that possible: on a document with a structure tree the correct reading order is known, so geometric output can be scored against it, and the resulting scorer then calibrates the geometry for untagged pages. That is the next piece of work, and it belongs in its own PR.

**State of this branch:** 1014 tests pass. On 404 real CVs, reversed words 20 against 24 before, affected files 7 against 11.

shlomsh:
This PR's premise, from `src/bidi_order.rs`'s module doc comment: PDF text-showing operators always paint left-to-right by glyph width, so RTL text must always be stored visually reversed, and `dominant_direction` + `visual_to_logical` should apply unconditionally once the document's direction reads RTL.

That doesn't hold. `TJ`'s numeric operand is subtracted from the pen position before the next glyph paints (ISO 32000-1 9.4.3), so a producer can store RTL text in ordinary logical order and use per-glyph negative advances to walk the pen leftward instead of reversing anything in the stream. A real government PDF does exactly this: [`rtl_logical_order_gov_form.pdf`](https://github.com/shlomsh/pdf-inspector/blob/bdba7b5/docs/evidence/rtl-visual-order/rtl_logical_order_gov_form.pdf) (redacted; arithmetic and extraction comparison in the [README](https://github.com/shlomsh/pdf-inspector/blob/bdba7b5/docs/evidence/rtl-visual-order/README.md)).

Extracting it: `main` gets it right today (0 words starting with a Hebrew word-final letter, 45 ending with one, correct spelling). This branch inverts that almost exactly (45/0, every word reversed), because `merge_text_items` reverses every item once `dominant_direction` reads RTL, with no per-item or per-producer escape hatch.

A minimal synthetic version of the same construct, no real document needed: [`rtl_logical_order.pdf`](https://github.com/shlomsh/pdf-inspector/blob/bdba7b5/tests/fixtures/rtl_logical_order.pdf) and `synthetic_rtl_pdf_pages_position_adjusted` in [`tests/integration_tests.rs`](https://github.com/shlomsh/pdf-inspector/blob/bdba7b5/tests/integration_tests.rs).

Separate observation, not the main point: `dominant_direction` is computed document-wide by design (the module comment argues per-line signal is too noisy), but its only production call site (`merge_text_items`, called per-page from `extract_page_text_items`) runs per page. A mostly-Hebrew document with one English-heavy page would flip direction on that page alone.

#303 takes a different approach: instead of deciding from "does this document contain RTL characters," it checks whether the extracted Hebrew is spelled correctly (final forms only at word end) before reversing anything, which is invariant to whichever layout mechanism the producer used. The two aren't mutually exclusive, the gate could decide *whether* to reorder and UAX #9 could still do the *how*, happy to sync on that if useful.

abimaelmartell:
This approach did not test good,

Appreciate the Effort!

🙏

PR #334, opened 2026-08-10
fix(rtl): restore logical order for RTL text at extraction time

SaarBarak:
## Problem

PDF content streams store text in *visual* order — glyphs appear in each show operation in the left-to-right sequence they are painted. For RTL scripts that means extracted text arrives character-reversed. The README already lists RTL support, but only line ordering was handled; the text inside each item stayed in paint order.

Taken from the first heading of a real Hebrew document (`pdf2md --raw`):

| | output |
|---|---|
| before | `**ןכדע**` … `##### רבמצד \| דחוימ םוסרפ 2020` |
| after | `**עדכן**` … `##### פרסום מיוחד \| דצמבר 2020` |

Note the second line recovers *word* order too, not just characters.

## Approach: restore at the item level

After extraction, item order within a line is already logical (RTL lines sort right-to-left), but each item's text is visual. Once items are concatenated the boundaries are lost, so no string-level pass downstream can be fully correct. The restoration therefore runs **per item, at extraction time**, which is the only level where the invariant is unconditional: a show operation is always in paint order.

`src/rtl.rs` applies the Unicode bidi algorithm's L2 reordering (UAX #9, via the `unicode-bidi` crate — zero-dependency, wasm-compatible). On the plain-text runs found in PDFs that mapping is an involution, so applying it to visual text with the original paragraph direction restores logical order. Direction is estimated per item by the same strict letter-majority rule the extractor already uses.

This is what makes mixed content work rather than naive reversal: digits, Latin runs, emails, phone numbers and URLs are level-2 runs that keep their internal order and land in the correct logical position. And because restoration precedes the serializer, all Markdown syntax (`###`, `**`, `<u>`, links, table pipes) is generated *around* already-logical text — no placeholder or escaping machinery anywhere.

Two refinements worth calling out:

- **Brackets are not mirrored.** ToUnicode maps mirrored glyphs back to logical characters, so positional reversal alone restores `(מיל׳)`. Applying rule L4 blindly (as `python-bidi` does) is what inverts them. A conservative repair pass — an unmatched closer *before* an unmatched opener of the same kind, a pattern well-formed text never produces — covers writers that store screen-form brackets instead.
- **Combining marks** (niqqud, harakat) stay attached to their base under reversal, while Hebrew punctuation interleaved with the points (maqaf, paseq, sof pasuq) is correctly *not* treated as combining. This is what fixes `רב־תחומי`.

`ActualText` replacements are excluded — the PDF spec defines them in logical order already — and the Arabic presentation-forms path, which already reversed at item level, is not double-processed.

## Table geometry

Three targeted fixes make tables RTL-aware:

- **Cell joins.** Cell fills that iterate items in ascending-X (paint) order now handle RTL, either by prepending RTL-majority fragments (`rtl::append_cell_text`, used by the line-based and heuristic builders, where it also reassembles split words) or by sorting cell items X-descending when the cell is RTL-majority (rect-grid and struct-tree builders).
- **Header recovery** sorts RTL header rows X-descending so multi-item header cells join in logical order.
- **Column order.** A final pass over the rendered Markdown (`rtl::restore_visual_order`) reverses per-row cell order in majority-RTL table blocks, so the logical first column comes first. It fires only when every row has the same cell count — an unescaped `|` in a math cell can then never scramble a row — and it honors `\|` escapes.

## Performance

Every entry point gates on an ASCII fast path plus one RTL-char scan, so LTR documents take identical code paths and allocate nothing extra.

All 27 fixture PDFs plus two English test documents — including a 300-page financial report — produce **byte-identical output** to the pre-patch binary. Exactly the 5 Hebrew-containing documents change.

Interleaved A/B runs on the 300-page English report put the patched binary within measurement noise of baseline. A control build with the restoration compiled out timed the same, which attributes the ~3% run-to-run spread to codegen and thermal noise rather than to this patch. Hebrew documents pay about 5% (1317 → 1380 ms on a 208-page document), only where the transformation actually applies.

## Validation

- `cargo test` — 1016 pass, 0 fail (860 unit including 15 new `rtl` tests, plus integration).
- `cargo clippy -- -D warnings` and `cargo fmt` clean.
- **Byte-identity gate** — the full corpus diffed against a saved baseline binary, as described above.
- **Phrase gate** — 14 known-correct logical phrases curated by reading the source documents (titles, bracketed ranks `(מיל׳)`, emails, phone numbers, multi-word table cells, tagged-table headers, `<u>`-wrapped headings, maqaf compounds) all appear verbatim in the new output.
- Output strictly improves on a `python-bidi` post-processing baseline: brackets correct where that inverted them, tab-separated lines in true logical order, table columns logicalized, and word order inside cells restored.

**One gap to flag:** CLAUDE.md asks for the `pdf-evals` regression suite before committing, and I don't have access to that sibling repo. The byte-identity gate above is my closest available substitute — it covers the same question (does anything change that shouldn't?) across every document I can run, but it is not the same corpus. Happy to rerun against `pdf-evals` if a maintainer can point me at it, or to have CI do it.

## Patch footprint

One new module, `src/rtl.rs` (416 lines, roughly half docs and 15 unit tests), plus 81 insertions / 42 deletions across 11 existing files — mostly one-line call-site swaps: `expand_ligatures(..)` → `normalize_show_text(..)` at the 5 show-text item constructors, `rtl::append_cell_text(..)` at 4 cell fills, an RTL-aware comparator in 2 cell sorts, one hook wrapping the Markdown return, `mod rtl;`, and one dependency line. No public API change — `mod rtl` is private and every function is `pub(crate)`.

## Known limitations

- Word equation objects extract with doubled glyphs (`𝑚𝑚`) from the source PDF itself, and math-symbol-heavy fragments count as LTR letters, so a few math/Hebrew bullet lines keep imperfect segment order.
- A writer that emitted RTL text in logical order per show-op *with normal LTR advances* would be double-reversed. No such writer is known — PDF pen advance makes that visually wrong in the viewer itself — and per-glyph positioned text is unaffected, since single-character items are no-ops.
- Scanned Hebrew still needs OCR; behavior there is unchanged.

---
## Summary by cubic
Restores logical reading order for RTL text at extraction time and makes lines/tables RTL-aware, and also recovers missing glyphs when ToUnicode CMaps have holes by consulting embedded font cmaps. LTR output stays byte-identical; U+FFFD and reversed segments in affected PDFs are fixed.

- **Bug Fixes**
  - RTL: Restore visual→logical per text item via `normalize_show_text` (Unicode bidi), skipping `ActualText` and Arabic presentation forms; keep combining marks with their base; no bracket mirroring, with a safe repair now run per item and per rendered line (per-cell and code-fence aware).
  - RTL container ordering: Decide reading order per container (line, cell, merged run), not per fragment:
    - `CellText` collects fragments in paint order and renders once complete; `join_run_fragments` reassembles split words before ordering; `order_cell_lines` orders each visual line inside a cell.
    - Line sort (`sort_line_items`) reverses RTL sequences then restores embedded LTR runs (equations like “x ∈ G” stay readable).
  - Tables: header recovery fills in paint order; rect/grid/struct/heuristic builders use `CellText` and per-line ordering; final Markdown pass `rtl::restore_visual_order` flips columns only for uniform, majority-RTL tables, honors `\|`, and skips fenced code blocks. Tagged (structure-tree) tables are pre-reversed via `positioned_struct_table`/`rtl::reverse_table_columns` so the final pass cancels exactly and preserves the tree’s logical column order.
  - CIDs with missing ToUnicode mappings: fill unmapped CIDs per code using the embedded font’s cmap when the descendant CIDFont declares glyphs (`/W` or `/CIDSet`) that the ToUnicode CMap can’t map; keep this “hole filler” out of whole-string scoring; fixes lone-CID operands that previously became U+FFFD; Identity-H and `/CIDToGIDMap /Identity` handled.
  - Performance/validation: ASCII fast paths; LTR path allocation-free and byte-identical; ~5% overhead only on affected RTL docs; embedded-font parse occurs only when a hole is detected (e.g. worst-case +1.4% on a hole-heavy file); new tests added for RTL and CID holes.

- **Dependencies**
  - Add `unicode-bidi` 0.3.

<sup>Written for commit 8f4df7dfc4491068c44725d15825b5045d0e1fbe. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/pdf-inspector/pull/334?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


SaarBarak:
Thanks — this was a useful review. Six of the seven are addressed in 36b15c6; the seventh I've left alone deliberately, with reasoning below.

Four of the findings turned out to be one root cause. The direction decision was being made from the *fragment* being added rather than from the *container* it was being added to, which only diverges when a run mixes directions. I reproduced the `[Hebrew][Latin][Hebrew]` case as a failing test first — it did produce `Hebrew Hebrew Latin` — and then fixed the cause rather than the sites:

| Finding | Resolution |
|---|---|
| `rtl.rs:122` — per-fragment cell assembly | `CellText` collects a cell's fragments in paint order and renders once complete; replaces `append_cell_text` at all five cell fills |
| `grid.rs:481` — header row pre-sorted descending-X | Fills in paint order and lets the cell decide direction |
| `detect_heuristic.rs:116` — per-fragment merge | `join_run_fragments`; touching pieces are joined into words first, so reordering can't split a word |
| `detect_rects.rs:1507` — whole-cell direction | `order_cell_lines` decides per visual line, so a Latin line under a Hebrew line is no longer reversed |
| `rtl.rs:91` — fenced code treated as a table | The pass now tracks code fences |
| `detect_struct.rs:468` — no coverage | Covered, along with the rest, by 8 new tests |

While fixing these I found a related defect the review didn't flag: reversing a sequence also reverses embedded left-to-right runs whose glyphs land in *separate* items. `x ∈ G` is painted with `x` leftmost, so the descending-X line sort emitted `G ∈ x`. No item-level pass can catch it — each item is a no-op alone and the run spans several. `order_rtl_sequence` now reverses the sequence and puts embedded LTR runs back in ascending order. Checked against the rendered page of a Hebrew number-theory summary, where one line carried three instances:

```
before  סדר של 𝐺 ∈ 𝑥 … כך ש 𝑒 = 𝑥. נסמן 𝑘 =)𝑥(𝑜𝑟𝑑
after   סדר של 𝑥 ∈ 𝐺 … כך ש 𝑥 = 𝑒. נסמן 𝑜𝑟𝑑(𝑥) = 𝑘
```

The bracket inversion resolved as a side effect: the brackets were logical all along and sat reversed because the run around them was.

**`text_utils.rs:203` (Arabic presentation forms) I've left as-is and documented instead.** Those items are already reversed by `expand_ligatures` via `reverse_visual_arabic`, so also running them through `restore_item_text` would reverse them twice — strictly worse for the common Arabic-only case. The real cost of the gate is narrower than described: only an item mixing presentation forms *with another RTL script* in a single show operation keeps the coarser reversal, which isn't something real writers produce. Happy to revisit if you disagree.

All 27 fixture PDFs and the English test document remain byte-identical to the pre-patch binary; 1025 tests pass, clippy and fmt clean.

SaarBarak:
Second round addressed in 93b3906. Two of these were right in a way the first round's checks couldn't have caught, and one I'm pushing back on with evidence.

**The performance point was the important one.** The container-order rewrite made every table cell allocate a `String` per fragment and the merger clone every fragment — costs the pre-RTL code never paid, on documents with no RTL at all. Interleaved A/B runs on a 300-page English report, rotating order, 15 rounds:

| build | median | min |
|---|---|---|
| `main` | 358 ms | 344 |
| PR, first revision | 363 ms | 356 |
| PR, container rewrite | 369 ms | 357 |
| **PR, this commit** | **360 ms** | 355 |

`CellText` now accumulates into a single `String` exactly as the original `push(' ')`/`push_str` pattern did, recording fragment offsets only from the first RTL fragment onward — everything before it is one opaque LTR chunk, since reversal plus the embedded-run pass restores its interior order anyway. A cell that never sees RTL never allocates the offset vector, `push` gates on `is_ascii` before character iteration, and `finish` consumes the accumulator so an LTR cell hands its string over by move. `join_run_fragments` borrows `&str` with a plain-append fast path.

**Fill order** (`tables/mod.rs`): correct, the slice is in stream order and `CellText` reconstructs reading order from paint order. RTL tables now sort ascending-X first — one index vector per table rather than per cell, and only when the table is RTL. It changed no output on the Hebrew corpus, so the exposure was latent rather than active, but the guarantee is now explicit.

**Fences** (`rtl.rs`): fixed to follow CommonMark — a fence closes only on the same delimiter character at a length at least as long, so a ``` block may contain `~~~` or a shorter backtick run.

**On digits as embedded-LTR-run content — I tried this and it regressed real output, so I've left the letters-only rule in place.** Two numbers that merely sit next to each other in an RTL sentence are separate runs: under UAX #9 the neutrals between them resolve to the paragraph level, so they must not swap. Counting digits merged them. A footnote in the corpus went from the correct

```
גאווה בהיותם ישראלים (יהודים – 92%). 98% מהערבים…
```

to

```
גאווה בהיותם ישראלים (יהודים – 98%). 92% מהערבים…
```

I rendered the page to confirm which one the document actually says — it's `92%` then `98%`. Leaving a split number reversed is the lesser error of the two, and the reasoning is now recorded at the predicate so it doesn't get "fixed" again later. If you'd like split numbers handled, it needs to key on fragment adjacency (touching, no word gap) rather than character class, which I'd rather do as a follow-up with corpus evidence than fold in here.

Two items from this round I have **not** addressed yet: the multi-line open-edge header in `detect_lines.rs:987`, and neutral-punctuation fragments in `join_run_fragments`. Both are real; I'd rather land them with corpus evidence than guess.

All 27 fixture PDFs remain byte-identical to the pre-patch binary. 1026 tests pass, clippy and fmt clean.

shlomsh:
Same root issue, flagging here too. This PR's module doc comment (`src/rtl.rs`) states content streams always store RTL text in visual order because "a show operation is always in paint order." Paint order being sequential doesn't imply left-to-right placement though, and `TJ`'s position adjustments (subtracted from the pen before the next glyph paints, ISO 32000-1 9.4.3) let a producer place glyphs right-to-left while keeping them in ordinary logical order in the stream.

A real document doing this: [`rtl_logical_order_gov_form.pdf`](https://github.com/shlomsh/pdf-inspector/blob/bdba7b5/docs/evidence/rtl-visual-order/rtl_logical_order_gov_form.pdf) (redacted; arithmetic in the [README](https://github.com/shlomsh/pdf-inspector/blob/bdba7b5/docs/evidence/rtl-visual-order/README.md)). `normalize_show_text` (`src/text_utils.rs`) gates `restore_item_text` only on `is_ascii()` and Arabic presentation forms, so this document's logical-order Hebrew gets reordered unconditionally and comes out character-reversed. `main` extracts it correctly today (0 words starting with a Hebrew word-final letter, 45 ending with one); this branch inverts that (45/0).

Minimal synthetic reproduction, no real document needed: [`rtl_logical_order.pdf`](https://github.com/shlomsh/pdf-inspector/blob/bdba7b5/tests/fixtures/rtl_logical_order.pdf) + `synthetic_rtl_pdf_pages_position_adjusted` in [`tests/integration_tests.rs`](https://github.com/shlomsh/pdf-inspector/blob/bdba7b5/tests/integration_tests.rs).

#303 handles this by gating the reversal on an orthographic check of the extracted text (Hebrew words spelled correctly, final forms only at word end) rather than reordering on "contains an RTL character" alone, which is invariant to how the producer laid the glyphs out. Might be worth adopting as a precondition here too, UAX #9 is still the better reordering algorithm once that decision is made. #303's own hand-rolled reversal mishandles non-ASCII Latin embedded in an RTL run (e.g. "François" round-trips as "oisçFran"), which `restore_item_text`'s bidi-based approach gets right.

PR #369, opened 2026-08-12
fix(xobjects): track text line matrix and handle T*/TL/'/"/Tc/Tw in Form XObjects

abimaelmartell:
## Problem

`https://www.nycourts.gov/reporter/files/bv/199AD3d.pdf` (NY Appellate Division Reports vol. 199, 1370 pages) extracts badly: we silently drop **79% of the text**, truncating mid-sentence:

```
Order, Supreme Court, New York County (Nancy M. Bannon,

[154 NYS3d 238] —        <- rest of the opinion simply gone
```

## Root cause

Every page of this PDF has a content stream that is just `q /I459 Do Q` — all real content lives in a Form XObject. That path goes through `src/extractor/xobjects.rs`, which is a **separate hand-rolled reimplementation** of the operator state machine in `content_stream.rs`. Nothing enforces parity between the two, and it had drifted badly:

| Issue | Effect |
|---|---|
| **No text line matrix (TLM)** | `Td`/`TD` were applied to the text matrix *already advanced by* `Tj`/`TJ`, so each line started where the previous one **ended**. Lines marched off the right edge and were dropped as off-page. |
| **`T*` entirely unhandled** | Used by **1331 of 1370** pages here. Never advanced the line. |
| `TL`, `'`, `"` missing; `TD` never set leading | Line advancement broken for these operators. |
| `Tc`/`Tw` hardcoded to `0.0` in width computation | Drifting advance widths, spurious spaces. |
| Text state not saved by `q`/`Q` | `Tc`/`Tw`/`TL`/`Tf` are part of the graphics state and must be restored. |

This is not an edge case: producers that emit `q /X Do Q` and put everything in a Form XObject are common in print-to-PDF and typesetting workflows (this file is PDFlib), so this parser is on the hot path for whole classes of real documents.

## Fix

Mirror `content_stream.rs`'s text-state handling in the form parser: a real `line_matrix`, the missing operators, `Tc`/`Tw` threaded into `compute_string_width_ts`, and text state saved/restored across `q`/`Q`.

## Results

Full 1370-page volume, word recall against a `pdftotext` reference:

| | words | recall |
|---|---|---|
| before | 116,304 | **19.2%** |
| after | 581,847 | **97.7%** |
| reference | 577,661 | — |

10-page subset, also scored against Mistral OCR:

| | chars | word recall | sequence similarity |
|---|---|---|---|
| before | 4,654 | 16.0% | 27.3% |
| **after** | 26,082 | **98.3%** | **97.7%** |
| Mistral OCR | 25,806 | 99.9% | 99.8% |
| pdftotext | 25,826 | — | — |

Body text now matches the reference word-for-word with italics preserved. Front matter recovered too — `ERIES 3d S` → `3d SERIES`, and the table of contents went from 2 truncated entries to all entries with page numbers.

## Also fixed: fill colour leaking across `q`/`Q`

A second bug in the same function, surfaced in review: a white fill set inside a `q`/`Q` pair leaked past the `Q`, so all following text was treated as invisible and dropped. `fill_is_white` is now saved and restored with the rest of the graphics state.

Isolating that commit against the previous one over all 203 eval PDFs, it changes **2 files**, both small improvements:

- `cambridge_excerpt` (−25 bytes) — drops a stray `**Success**` that was being injected into the middle of a sentence
- `17b8a322802e_EAT-Lancet…Russian` (−42 bytes) — heading/body reading order: `Масштабный переход к устойчивым рационам` is now one heading instead of fusing with body text (`устойчивым рационамуровнем доходов`)

<sub>Correction: commit `1618387`'s message credits this fix with the large recoveries listed below (cambridge +8.5KB, MTU +7.4KB, etc.). That attribution is wrong — those all come from the first commit's text-line-matrix work. Measured per-commit numbers are as stated here. The commit message is left as-is rather than rewriting pushed history.</sub>

## Verification

- `cargo fmt` clean, `cargo clippy -- -D warnings` clean.
- **1,045 tests pass**, including **8 new** unit tests covering line-matrix-relative `Td`, `T*`, `TD` setting leading, `'`, `"`, `Tc` advance widths, and `q`/`Q` restore of both text state and fill colour — each driven through a page whose content stream is only `q /X1 Do Q`. The fill-colour test was confirmed to fail without the fix.

For the regression check I ran **both binaries (`origin/main` vs this branch) directly over all 203 eval PDFs and diffed the outputs**, rather than relying on `bench.py test`. That command defaults to `--update`, so it rewrites the very snapshots it compares against — sequential runs measure against a moving baseline. The direct A/B is snapshot-independent:

| | count |
|---|---|
| PDFs compared | 203 |
| byte-identical output | **188** |
| output differs | 15 |
| extraction fails in **both** (pre-existing, unrelated) | 9 |

All 15 differences were inspected individually. **None is a regression:**

| PDF | Δ bytes | Change |
|---|---|---|
| `cambridge_excerpt_9781009122733` | +8,562 | recovers disclaimer + syllabus paragraphs dropped entirely before |
| `MTUAeroEnginesAG_AnnualReport_2024` | +7,429 | table column order fixed (`01\|02\|03\|04\|05` was `01\|02\|03\|05 04`) |
| `2025_findings-acl_668` | +1,370 | recovered text |
| `HTM_02-01_Part_A` | +1,309 | recovers "validation and verification", "Gateway Ref:" |
| `HuttoISDWorkPerks` | −172 | **exploded per-character table → clean prose** |
| `ebgt7isj04ophcq` | −3,020 | **exploded per-character table → clean prose** |
| `BlueDiamond_AlmondFacts` | +1 | column reading order fixed (was interleaving two columns mid-sentence) |
| `dergi_0020_04` | 0 | author names correctly grouped on one line |
| `The_Times_March_2025` | −2 | `Monda y M arch 3` → `Monday March 3` |
| `Soil_Quality-Earthworms` | −3 | `0 C T-W C NT-WC NT-S C` → `0 CT-WC NT-WC NT-SC` |
| `IN_Annual_Report_2017` | −1 | table cell split correctly |
| `2103_07786` | −9 | drops a stray `] )` artifact, adds missing content |
| `17b8a322802e_EAT-Lancet…Russian` | −49 | reading order fixed: `Говядина, баранина, свинина` (pork was misplaced before beef/lamb) |
| `Stijn_SB_doc_-_NYCempty_commercial` | 0 | chart axis tick labels regrouped (cosmetic) |
| `140120260213559638` | −69 | Japanese slide deck; fewer spurious spaces (`35点` vs `35 点`) |

The two files that got *smaller* are the biggest wins — the byte count drops because garbled per-character table markup is replaced by readable prose.

## Not addressed (pre-existing, out of scope here)

Both are independent of this change and would churn a large share of the ~200 eval snapshots, so they're left as follow-ups:

1. **False-positive tables (56/1370 pages).** The front-matter justices list still comes out scrambled (`|OLANDO|COSTA|NIL|INGH||`). Extraction is now correct there — verified all 9 items on that baseline with correct x/width; small caps are drawn as alternating 9.98pt/6.74pt runs, and the true column gap (72pt) is cleanly separable from the small-caps sub-gaps (0–3.3pt). Purely downstream table misdetection.
2. **Line-break hyphenation not rejoined** (`compensa` + `tion`, `de` + `fendant`). This accounts for essentially the entire remaining 98.3% → 99.9% gap vs Mistral. No cross-line dehyphenation exists in the codebase today. Worth doing, but it has a real downside: genuine compounds broken at the hyphen become `sixmonth` (pdftotext gets that wrong too).

### DIFF
--- a/src/extractor/xobjects.rs
+++ b/src/extractor/xobjects.rs
@@ -246,19 +246,49 @@ fn extract_form_xobject_text_inner(
     let mut current_font = String::new();
     let mut current_font_size: f32 = 12.0;
     let mut text_matrix = [1.0f32, 0.0, 0.0, 1.0, 0.0, 0.0];
+    // Text line matrix (TLM) — Td/TD/T* move relative to the start of the
+    // current line, not to the position left by the last show operator.
+    let mut line_matrix = [1.0f32, 0.0, 0.0, 1.0, 0.0, 0.0];
+    let mut text_leading: f32 = 0.0; // TL parameter (text-space units)
+    let mut char_spacing: f32 = 0.0; // Tc parameter
+    let mut word_spacing: f32 = 0.0; // Tw parameter
     let mut in_text_block = false;
     let mut fill_is_white = false;
     let mut ctm = base_ctm;
-    let mut ctm_stack: Vec<[f32; 6]> = Vec::new();
+
+    // Text state (Tc/Tw/TL/Tf) is part of the graphics state and must be
+    // saved/restored by q/Q alongside the CTM.
+    #[derive(Clone)]
+    struct GraphicsState {
+        ctm: [f32; 6],
+        char_spacing: f32,
+        word_spacing: f32,
+        text_leading: f32,
+        current_font: String,
+        current_font_size: f32,
+    }
+    let mut ctm_stack: Vec<GraphicsState> = Vec::new();
 
     for op in &content.operations {
         match op.operator.as_str() {
             "q" => {
-                ctm_stack.push(ctm);
+                ctm_stack.push(GraphicsState {
+                    ctm,
+                    char_spacing,
+                    word_spacing,
+                    text_leading,
+                    current_font: current_font.clone(),
+                    current_font_size,
+                });
             }
             "Q" => {
                 if let Some(saved) = ctm_stack.pop() {
-                    ctm = saved;
+                    ctm = saved.ctm;
+                    char_spacing = saved.char_spacing;
+                    word_spacing = saved.word_spacing;
+                    text_leading = saved.text_leading;
+                    current_font = saved.current_font;
+                    current_font_size = saved.current_font_size;
                 }
             }
             "cm" => {
@@ -321,6 +351,7 @@ fn extract_form_xobject_text_inner(
             "BT" => {
                 in_text_block = true;
                 text_matrix = [1.0, 0.0, 0.0, 1.0, 0.0, 0.0];
+                line_matrix = text_matrix;
             }
             "ET" => {
                 in_text_block = false;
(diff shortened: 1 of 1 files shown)


abimaelmartell:
Eval snapshots for this change: firecrawl/pdf-evals#68 (draft until this merges) — 14 markdown snapshots + 6 metadata files.

abimaelmartell:
Follow-up stacked on this branch: #371 — merges small-caps runs so they stop reading as table columns (fixes the garbled justices list noted under "Not addressed" above). Drops false-positive tables on 199AD3d.pdf from 65 to 52.

PR #378, opened 2026-08-13
Restore logical reading order for visual-order Hebrew

shaydavid9:
## The problem

Hebrew PDFs that store glyphs in visual order extract reversed, and nothing downstream catches it:

```
## מ"עב ןואג ריש          ← should be: שיר גאון בע"מ
|כ"הס|ריחמ|טוריפ|תומכ|    ← should be: |כמות|פירוט|מחיר|סה"כ|
```

The output is not merely untidy — it is unreadable, and an agent consuming it has no way to tell.

## Why it happens

The reversal already exists. `expand_ligatures()` calls `reverse_visual_arabic()` whenever it sees visual-order text. The trigger is the catch:

```rust
let had_presentation_forms = text.chars().any(is_arabic_presentation_form);
```

Arabic presentation forms (U+FB50-FDFF, U+FE70-FEFE) are an Arabic-only phenomenon. Visual-order Hebrew is written with plain base codepoints (U+05D0-05EA), so this is always `false` and the reversal never runs. Hebrew falls through a gap: `is_rtl_char` knows about it, `sort_line_items` sorts for it, but nothing ever flips it.

## The signal

Hebrew has no presentation-form marker, so ordering has to come from word shape. Its five final forms — ך ם ן ף ץ — are legal **only** as a word's last letter. When a producer lays glyphs out visually, extraction walks each word backwards and those letters surface at the *start* of words instead.

`is_visual_order_hebrew()` counts which end they land on. Ties — including text with no final forms at all — report logical, so ambiguous input is left alone rather than scrambled. Guessing wrong in that direction is the expensive mistake: it corrupts a correct document.

## Where the decision is made

Once per document, in `clean_markdown()` — not per text item. A producer picks one ordering for the whole file, and `expand_ligatures()` only ever sees fragments, most too short to carry the signal. Deciding per fragment would flip some lines of a page and not others, which is worse than either consistent answer.

Markdown table rows also get their cell order reversed: an RTL table's leftmost visual column is its last logical one.

## Notes for review

- `reverse_visual_arabic` → `reverse_visual_rtl`. The function was already script-agnostic; it now has two callers.
- **Arabic is deliberately untouched.** It keeps its existing per-item path and is explicitly out of scope for the document pass, so it is never reversed twice.
- Opt out with `MarkdownOptions.fix_rtl_order`.
- Two rough edges belong to the existing reversal and are left alone to keep this diff focused: mirrored pairs are not flipped (`)ח.פ(`), and a comma adjacent to a number stays glued to it (`,12`). Both already affect Arabic today; a follow-up could route the reversal through a full UBA implementation.

## Verification

`cargo fmt`, `cargo clippy` (no new warnings from these files) and `cargo test` all pass — 923 unit + 162 integration, including 15 new tests.

Behaviour was diffed against `main` with a release build:

| Corpus | Result |
|---|---|
| All 27 `tests/fixtures/*.pdf` | byte-identical |
| 10 non-Hebrew documents | byte-identical |
| 29 Hebrew documents | 24 visual → now correct; **5 already logical → untouched** |

No false positives — the "already logical, left alone" column is the one that matters, and it is clean.

I could not run the `pdf-evals` snapshot suite (sibling repo, not public as far as I can tell), so that check is outstanding and worth running before merge. The test corpus above is real-world Hebrew paperwork — invoices, payslips, bank transfers — which I can't attach as fixtures since it is personal; the unit tests use synthetic strings only. Happy to add a synthetic Hebrew fixture PDF if you want integration coverage.

---
## Summary by cubic
Restores logical reading order for visual-order Hebrew so extracted markdown is readable. Previously, Hebrew text reversed on extraction, Arabic in mixed docs could be reversed twice, and inline markup could invert; now the pass is scoped to Hebrew and preserves inline tags and emphasis.

**Key points**
- Detect visual-order Hebrew via final-form letters: count appearances at word starts vs ends; ties report logical and leave text unchanged.
- Decide once per document in `clean_markdown()`; scope line reversal to lines with Hebrew and no Arabic so Arabic is never double-reversed.
- Reverse prose with `reverse_visual_rtl` (renamed from `reverse_visual_arabic`); preserve heading/list/quote markers.
- Reverse pipe-delimited table cell order; separator rows (`|---|`) stay unchanged.
- Preserve inline markup: keep `<u>`, `*`/`**`, and `` ` `` runs in place and reverse only text segments.
- Add `MarkdownOptions.fix_rtl_order` (default true); disable to keep visual order. Non-Hebrew docs remain byte-identical; already-logical Hebrew stays untouched; tests updated and pass.

<sup>Written for commit dc7a11fac8fcff23e9df967e9a4e0562d84bb7b2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/pdf-inspector/pull/378?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/markdown/mod.rs
+++ b/src/markdown/mod.rs
@@ -951,6 +951,8 @@ pub struct MarkdownOptions {
     pub include_page_numbers: bool,
     /// Strip repeated headers/footers that appear on many pages
     pub strip_headers_footers: bool,
+    /// Restore logical reading order for Hebrew stored in visual order
+    pub fix_rtl_order: bool,
 }
 
 impl Default for MarkdownOptions {
@@ -980,6 +982,9 @@ impl Default for MarkdownOptions {
             include_links: true,
             include_page_numbers: false,
             strip_headers_footers: true,
+            // On by default: visual-order Hebrew is unreadable and silently
+            // wrong, and the detector is a no-op on every other document.
+            fix_rtl_order: true,
         }
     }
 }
--- a/src/markdown/postprocess.rs
+++ b/src/markdown/postprocess.rs
@@ -4,7 +4,8 @@ use regex::Regex;
 
 use super::{MarkdownOptions, MarkdownProfile};
 use crate::text_utils::{
-    is_page_number_line, is_rtl_char, is_visual_order_hebrew, reverse_visual_rtl,
+    is_arabic_char, is_hebrew_letter, is_page_number_line, is_visual_order_hebrew,
+    reverse_visual_rtl,
 };
 
 /// Clean up markdown output with post-processing
@@ -66,8 +67,8 @@ pub(crate) fn clean_markdown(mut text: String, options: &MarkdownOptions) -> Str
 /// Documents already in logical order, and documents with no Hebrew, are
 /// returned untouched.
 ///
-/// Arabic is deliberately out of scope here: it is already restored per text
-/// item during extraction, and reversing it a second time would undo that.
+/// Arabic is out of scope: it is already restored per text item during
+/// extraction, so [`restore_rtl_line`] skips any line carrying it.
 fn restore_rtl_reading_order(text: &str) -> String {
     if !is_visual_order_hebrew(text) {
         return text.to_string();
@@ -85,7 +86,11 @@ fn restore_rtl_reading_order(text: &str) -> String {
 
 /// Restore one line, keeping markdown structure in place around the prose.
 fn restore_rtl_line(line: &str) -> String {
-    if !line.chars().any(is_rtl_char) {
+    // Scoped to Hebrew, not RTL generally. Arabic reaches this point already
+    // in logical order — `expand_ligatures` restored it per text item during
+    // extraction — so reversing it here would undo that work. A line carrying
+    // both scripts is genuinely ambiguous, and is left alone.
+    if !line.chars().any(is_hebrew_letter) || line.chars().any(is_arabic_char) {
         return line.to_string();
     }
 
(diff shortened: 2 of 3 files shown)


shaydavid9:
Thanks — the Arabic double-reversal was real, and I've fixed it in dc7a11f.

I reproduced it before changing anything: with a Hebrew line supplying the document-level signal, `مرحبا` came back as `ابحرم`. `restore_rtl_line` gated on `is_rtl_char`, which matches Arabic too, so every Arabic line in such a document was reversed a second time — undoing what `expand_ligatures` had already done per item. The docstring asserted the opposite; the code never backed it up.

It now gates on Hebrew letters and skips any line that also carries Arabic, since a line with both scripts can't be judged from a single signal.

The `<u>` finding was also real — `<u>שלום עולם</u>` was coming out as `/u><שלום עולם<u>`. `reverse_segments()` now splits on inline markup and reorders only the text between, covering emphasis and code spans too.

On the niqqud report: I couldn't reproduce a failure. Vocalized `םָלוֹע םוֹלׁש` is still detected as visual, because enough unvocalized words carry the signal. Worth noting the failure mode is safe either way — losing the signal means the text is left untouched, not scrambled. Happy to make word-splitting mark-aware anyway if you'd prefer the belt and braces.

Three regression tests added, one per defect. Full suite green at 926 unit + 162 integration, and I re-ran the corpus comparison: fixtures and non-Hebrew documents still byte-identical, 24 Hebrew documents fixed, 5 already-logical ones untouched.

abimaelmartell:
This was fixed in #440

Appreciate the help!

PR #387, opened 2026-08-14
fix(cli): escape JSON error output

NgoQuocViet2001:
### What

The `--json` exit in `pdf2md`'s main conversion path interpolates the error straight into its object:

```rust
Err(e) => {
    if json_output {
        println!(r#"{{"error":"{}"}}"#, e);   // no escaping
```

A `PdfError` message carrying a quote or a newline produces JSON that no consumer can parse. `missing required dictionary key "Pages"` is the everyday case:

```
before  {"error":"PDF parsing error: missing required dictionary key "Pages"
while loading catalog"}          <- unparseable

after   {"error":"PDF parsing error: missing required dictionary key \"Pages\"\nwhile loading catalog"}
```

### How

Both `--json` error exits now go through a small `format_error_json`, placed beside and mirroring the existing `format_ocr_error_json`:

```rust
fn format_error_json(error: &str) -> String {
    format!(r#"{{"error":"{}"}}"#, json_escape(error))
}
```

The items-json exit was already calling `json_escape` inline, so routing it through the helper is not a behaviour change there — it just means the two paths cannot drift, and the emitted object is testable rather than only the escaper.

I did not give this one a `schema_version`, unlike the OCR envelope: that would change the shape of an output that already ships. Happy to add it if you'd rather the two converge.

### Rebased and reshaped

This PR was opened before #434 and had gone stale — resetting onto current `main` resolved the conflict, and it turned out the original patch was aimed at a site you have since fixed. The items-json path escapes on `main` today; the conversion path is the one still open. So this is now a much smaller change against that remaining site rather than a rebase of the original diff.

### Test plan

- Ran: `cargo test --bin pdf2md` — **4 passed, 0 failed**, including the existing `ocr_json_errors_use_the_same_versioned_envelope`.
- New: `convert_json_errors_are_escaped` asserts the full emitted object for a message containing both a quote and a newline.
- Ran: `cargo clippy --bin pdf2md` — no warnings.
- Ran: `rustfmt --check src/bin/pdf2md.rs` — clean.

**Not run:** `cargo fmt --all -- --check` and the full `cargo test`. `cargo fmt` across this workspace overflows its stack on my Windows machine — it does so on unmodified `main` too, so it is my environment rather than this change — which is why I checked the one touched file with `rustfmt` directly. Flagging it rather than letting CI be the first to find out. An earlier revision of this description claimed both had been run; that was carried over from the original August 14 version of this PR and is not accurate for this commit.


NgoQuocViet2001:
Merged current `main` in (the branch was 11 behind; it merged clean, no conflict) and re-ran the Clippy job's two steps verbatim against the merged tree:

```
cargo clippy -- -D warnings                  -> clean
cargo clippy --features ocr -- -D warnings   -> Finished, no warnings
cargo test --bin pdf2md                      -> 4 passed, 0 failed
rustfmt --check src/bin/pdf2md.rs            -> clean
```

Both are green here, so I expect the re-run on this head to be too. The diff is unchanged in scope — still `src/bin/pdf2md.rs` only, +25/-3.

Being straight about what I do and don't know: I could not reproduce the failure, and I can't read the job log without a token, so I have not identified a cause in this change. The red run finished in 24 seconds, which is short enough that it may not have got as far as linting. If it reproduces on this head, paste the failing lint and I'll fix it properly rather than keep guessing.

One thing I did turn up while looking, unrelated to this PR and worth only what it's worth: `cargo clippy --all-targets` (which the Clippy job does not pass) fails on `main` with 22 `useless_vec` errors, all in the `#[cfg(test)]` module of `src/tables/grid.rs`:

```
error: useless use of `vec!`
   --> src/tables/grid.rs:697:26
    |
697 |         let items_data = vec![make_item("A", 100.0, 500.0, 10.0)];
    |                          help: you can use an array directly
```

Pre-existing and outside this change, so I have left it alone — mentioning it in case you ever want to widen that job's target selection.

PR #396, opened 2026-08-15
fix: route scan-backed narrow text pages to OCR

earayu:
## Summary

- route a page to OCR when one displayed image covers at least 80% of the inherited CropBox/MediaBox and the page has no non-empty text item at least 2pt wide
- reuse the existing `scanned` OCR reason and the already extracted page items
- calculate image coverage in the same unrotated PDF user space as the normalized page box

## Regression coverage

- adds the public `natural-pdf` fixture `24640152.pdf` at commit `99593235bb2f75467ae5bb90e6e478f00840caa7`
- verifies its remaining qualifying pages (9, 11, and 12) are routed to OCR with empty unreliable Markdown
- verifies the exact 80% image boundary, the 2pt text boundary, and a 79% image negative case
- mutation checks confirmed that removing either boundary makes the corresponding negative regression fail

Fixture SHA-256: `cedb9872a06143f275fd88647cd9aa988299d686bead0b43a747b557f9e9bae4`

## Validation

- `cargo fmt --all -- --check`
- `cargo test` (948 library, 2 CLI, 163 integration, and 2 doc tests)
- `python3 -m unittest discover -s scripts/tests` (18 tests)
- `python3 scripts/version.py --check`
- `maturin build --release`; installed the wheel into a clean Python 3.11 virtual environment and verified `extract_pages_markdown()` returns pages 9, 11, and 12 with reason `scanned`

This change intentionally does not bump the package version; the repository's release workflow can publish the next version after merge.

---
## Summary by cubic
Routes scan-backed pages with only sub‑2pt text to OCR. Previously these pages stayed native and published unreliable text; now they get OCR with reason "scanned" and publish empty markdown.

- Detects pages where one displayed image covers at least 80% of the inherited CropBox/MediaBox and no non-empty text item is at least 2pt wide.
- Computes coverage in unrotated PDF user space against the normalized page box; handles rotated pages.
- Adds a `needs_ocr` condition and reuses the existing "scanned" reason; `extractor::get_page_box` is now `pub(crate)`.
- Extends tests with the `natural-pdf` fixture `tests/fixtures/natural-24640152.pdf`; verifies pages 9, 11, and 12 route to OCR, plus exact 80%/2pt boundaries and a 79% negative case.

<sup>Written for commit b9385c9f6af52c4985df08653cca85c770600c86. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/pdf-inspector/pull/396?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


earayu:
REQUEST CHANGES: the new coverage check does not always compare the image and page box in the same coordinate space.

On exact head `b9385c9f6af52c4985df08653cca85c770600c86`, I built a one-page PDF with an inherited offset CropBox `[10 20 110 220]`, one image exactly filling that 100 x 200 visible box, and two 0.1pt rotated text runs. The text runs trigger `correct_rotated_page`, which changes the image item from `(10,20,100,200)` to `(20,-10,100,200)`. `page_has_scan_backed_no_substantial_text` then intersects that transformed item with the still-unrotated CropBox, reducing actual 100% coverage to 76.5%. The production API returns `pages_needing_ocr=[]`; the contract requires page 1 with reason `scanned`.

Root cause: `correct_rotated_page` rewrites every `TextItem`'s x/y but leaves image width/height unchanged, while `get_page_box` remains in raw PDF user space. Swapping width/height alone is insufficient because an offset CropBox also exposes the origin mismatch.

Please keep the image bbox used by this rule and the inherited normalized CropBox/MediaBox in the same unrotated PDF user space, without an extra content-stream pass, and add a production-path regression covering inherited/offset CropBox + content-stream rotation + full-page image. The regression should assert `pages_needing_ocr`, `ocr_reasons_by_page`, and per-page `needs_ocr`/`ocr_reason` together.

Other evidence on this exact head: `cargo test` passed (948 unit + 163 integration + 2 bin + 2 doc), format/Clippy/version/developer-script checks passed, and a built abi3 wheel returned `[9,11,12]` with `scanned` consistently for the natural fixture. The 2.2 MiB fixture is excluded from the published crate by the existing Cargo include allowlist, so I found no packaging-size blocker.

earayu:
Closing this upstream proposal. We will keep the behavior change inside aperag-enterprise instead, so no pdf-inspector release or dependency update is needed.

PR #408, opened 2026-08-17
fix: extract_pages_markdown no longer splits a word with a mid-word <u> tag

### LINKED ISSUE #397: extract_pages_markdown emits <u> tags that split a word when underline flag differs across positioned items in the same visual word
## Summary

A display heading laid out as several positioned text items can carry the `is_underline` flag on only one of those items. `extract_pages_markdown` (and `process_pdf`) then wraps just that item in `<u>...</u>`, splitting a single visual word with a tag boundary:

```
The Institutional <u>We</u>alth Landscape
```

Rendered, `<u>` never opens or closes on a non-word boundary, so this isn't formatting that was in the source — it's an artifact of how the layout engine grouped (or failed to group) items into a word before checking the underline flag.

## Environment

- pdf-inspector (Python binding) via `pip install pdf-inspector`
- Reproduced via `pdf_inspector.process_pdf()` / `extract_pages_markdown()`

## Reproduction

On an 11-page report exported from Chrome ("Save as PDF", producer `Skia/PDF`), page 1's title is `The Institutional Wealth Landscape`, styled as a single 42pt heading.

```python
import pdf_inspector
result = pdf_inspector.process_pdf("report.pdf")
print(result.markdown)
# ...
# # The Institutional <u>We</u>alth Landscape
```

Checking the underlying positioned items confirms the split happens at the item level:

```python
items = pdf_inspector.extract_text_with_positions("report.pdf")
for t in items:
    if t.text in ("The Institutional", "We", "alth Landscape"):
        print(t.text, t.is_underline, t.x, t.font_size)
# 'The Institutional'   is_underline=False   x=45.0   size=42.0
# 'We'                  is_underline=True    x=45.0   size=42.0
# 'alth Landscape'      is_underline=False   x=101.8  size=42.0
```

Three items, contiguous, same font size, same baseline — clearly one visual word ("Wealth") split across two content-stream text-showing operators, with `is_underline` true on only the first fragment. Six other headings in the same document exhibit the same pattern (`<u>01 Geopolitical and macro risk</u>`, `<u>Master source index</u>`, etc.) — those happen to be whole-phrase, so they render fine, but the underlying cause is the same per-item flag not being reconciled before Markdown conversion.

## Expected behavior

Either:
1. The word-grouping step that assembles positioned items into `TextLine`/word units should also reconcile `is_underline` across items that form one word (e.g. before word-boundary characters, or via majority/first-non-trivial fragment), so the tag wraps the whole word or none of it, or
2. If the items are being treated as separate underline runs on purpose, the Markdown converter should not open/close a `<u>` tag on a non-word-boundary character.

## Impact

Cosmetic on its own, but it corrupts every downstream consumer that treats the Markdown as ground truth text — search indexing, embeddings, and diffing all see `Wealth` broken into two tokens (`We`, `alth`) with an HTML tag between them. In our case it also required a post-processing repair step before the Markdown could be treated as reliable extraction output.

## Suggested severity

Low-to-medium: doesn't affect classification or table extraction, but it's silent — nothing in the output flags the corruption, so it can only be caught by comparing against the source.

---

Happy to share the source PDF (Skia/PDF producer) privately if it helps reproduce — it isn't public but I can send a minimal repro extracted from it.

MADENIYOU:
Fixes #397.

## Root cause

`TextLine::text_with_formatting` (src/types.rs) toggles the `<u>` tag per positioned item, using each item's own `is_underline` flag independently. A single visual word can be split across several content-stream text-showing operators, and the source PDF can carry the underline flag on only one fragment — e.g. an underline rectangle whose geometry happens to cover only part of a word's width, a rendering artifact rather than intentional partial-word styling.

Layout-level merging (`merge_text_items`) already refuses to combine items across a style difference (its own comment explains why: "OR-ing underline instead would stretch `<u>` spans over neighboring plain text"), so such fragments arrive at `TextLine::text_with_formatting` as adjacent, unmerged items with zero gap between them. Each toggling its own tag independently produces `<u>We</u>alth` for a word like "Wealth" — not formatting that was in the source.

## Fix

Don't toggle the underline tag between two items when both:

1. Are **truly geometrically contiguous** — gap under 2% of font size. This is intentionally much tighter than the general prose-spacing threshold (`needs_space_between`), because the goal here is different: distinguishing "this is the same word, split at the extraction layer" from "these are two separate items a normal layout heuristic happens to join without inserting a space" (many unrelated reasons exist for the latter — punctuation attachment, hyphenation, kerning).
2. Are **not separated by closing punctuation** — a trailing `.`/`,`/`;`/etc. isn't part of the word before it, even when kerned tightly against it. This condition was needed after a geometry-only version regressed existing snapshot output (`td9264.pdf`, a real, tightly-kerned legal document) where `<u>` legitimately opens/closes right at a punctuation boundary.

A continuation meeting both conditions now inherits whichever underline state is already open, instead of evaluating its own (possibly spurious) flag — so the tag wraps the whole word (or none of it), never splitting mid-word.

## Testing

- New unit test `underline_does_not_toggle_mid_word`, reproducing the issue's exact pattern (a word split across 4 items, underline flagged on only one interior fragment). Verified it fails without the fix (reproduces the exact reported `<u>We</u>alth` split) and passes with it.
- New unit test `underline_still_toggles_at_a_real_word_boundary`, confirming the fix doesn't suppress underline at genuine word boundaries.
- Iterated against the existing snapshot suite: an early geometry-only version of this fix regressed `td9264.pdf` (extended `<u>` across sentence punctuation and into unrelated following prose); the punctuation condition above was added specifically to fix that regression, and the full suite now passes unchanged.
- `cargo fmt`, `cargo test` (all passing, 2 new), `cargo clippy --all-targets -- -D warnings` clean.

---
## Summary by cubic
Prevents `extract_pages_markdown` from opening/closing `<u>` mid-word. Previously underline toggled per item and could emit `<u>We</u>alth`; now tightly contiguous word fragments are grouped and one underline decision is applied, avoiding order dependence and spans across real word boundaries.

- Precomputes effective underline via `TextLine::resolve_underline_by_word_group`: groups only when the gap is under 2% of font size, is direction-aware (RTL), agrees with the render loop’s spacing decision (`should_join_items`), preserves the hyphen no-space exception from `needs_space_between`, excludes closing punctuation on either side, treats strikeout as a hard boundary, and does not treat `'` as a boundary.
- Resolves a run’s underline by majority character count (ties favor plain text); the render loop uses this effective flag. Strikeout still overrides underline; bold/italic behavior unchanged. Tests cover hyphenated joins, order-independence, RTL contiguity, strikeout isolation, symmetric punctuation checks, apostrophe-in-word contractions, CID-font spacing alignment, and the original mid-word and real boundary cases.

<sup>Written for commit ace4153a652f66979f42a272642c5b6f9c42ffc4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/pdf-inspector/pull/408?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -164,6 +164,12 @@ impl TextLine {
 
         let single_char_threshold = self.adaptive_threshold;
 
+        // Resolve each item's *effective* underline up front, one run of
+        // tightly-continuous fragments at a time, rather than toggling the
+        // `<u>` tag per item as the render loop walks through. See
+        // resolve_underline_by_word_group's doc comment for why.
+        let effective_underline = self.resolve_underline_by_word_group(format_decorations);
+
         let mut result = String::new();
         let mut current_bold = false;
         let mut current_italic = false;
@@ -194,43 +200,6 @@ impl TextLine {
             // we push text_trimmed below (which strips it).
             let has_leading_space = text.starts_with(' ');
 
-            // True geometric contiguity with the previous item — tight
-            // enough that these two items are the same visual word split
-            // across positioned-text fragments, not merely two items
-            // `needs_space_between` happens to join without inserting a
-            // space. Some real documents kern punctuation just as tightly
-            // against the preceding word as a genuine same-word split, so
-            // geometry alone isn't enough (see the punctuation check
-            // below); this narrows out most unrelated no-space joins
-            // first (larger, deliberate letter-spacing gaps, real word
-            // gaps, etc.).
-            let is_tight_geometric_continuation = i > 0 && {
-                let prev_item = &self.items[i - 1];
-                prev_item.width > 0.0 && {
-                    let gap = item.x - (prev_item.x + prev_item.width);
-                    gap.abs() < prev_item.font_size * 0.02
-                }
-            };
-
-            // A trailing period/comma/etc. isn't part of the word before
-            // it, even when kerned touching it — closing punctuation is a
-            // legitimate style boundary regardless of geometric gap.
-            let curr_starts_with_join_punct = text_trimmed
-                .chars()
-                .next()
-                .is_some_and(|c| matches!(c, '.' | ',' | ';' | '!' | '?' | ')' | ']' | '}' | '\''));
-
-            // Whether a space will separate this item from the previously
-            // rendered content — i.e. whether this item starts a new word
-            // rather than continuing the previous item's word with zero gap.
-            let is_word_boundary = i == 0
-                || result.is_empty()
-                || result.ends_with(' ')
-                || needs_space
-                || has_leading_space
-                || !is_tight_geometric_continuation
-                || curr_starts_with_join_punct;
-
             // Check for style changes. Source decorations are exclusive:
             // `<u>`/`<s>` content stays free of `**`/`*` markers — consumers
             // (and the eval harnesses this feeds) match tag content literally,
@@ -238,26 +207,7 @@ impl TextLine {
             // emitted as struck text because deletion is the stronger semantic
             // distinction in redline documents.
             let item_strikeout = format_decorations && item.is_strikeout;
-            let item_underline_raw = format_decorations && item.is_underline && !item_strikeout;
-            // Never toggle the underline tag on a non-word-boundary
-            // character. A single visual word can be split across several
-            // positioned items at the content-stream level, and the
-            // source PDF can carry the underline flag on only one
-            // fragment (e.g. an underline rectangle covering just the
-            // first fragment's width) — layout-level merging already
-            // refuses to combine such items across a style difference
-            // (see extractor::mod's merge_text_items), so they arrive
-            // here as adjacent, unmerged items with zero gap between
-            // them. Toggling `<u>` per-item in that case opens/closes the
-            // tag mid-word (`<u>We</u>alth`), which isn't formatting that
-            // was in the source. A same-word continuation instead
-            // inherits whichever underline state is already open. See
-            // #397.
-            let item_underline = if is_word_boundary {
-                item_underline_raw
-            } else {
-                current_underline
-            };
+            let item_underline = effective_underline[i] && !item_strikeout;
             let item_bold = format_bold && item.is_bold && !item_underline && !item_strikeout;
             let item_italic = format_italic && item.is_italic && !item_underline && !item_strikeout;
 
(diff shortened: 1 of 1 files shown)


MADENIYOU:
All five confirmed and fixed — thanks, this was a thorough round.

**Order-dependence (P2)**: Right — serial inheritance made the result depend on fragment order, not just source formatting. Replaced with a two-pass design: group tightly-continuous items into runs first, then decide each run's underline status once via majority character count (ties favor not-underlined). Same source pattern now always resolves identically regardless of which fragment carries the flag.

**RTL gap (P2)**: Confirmed, the gap calc was LTR-only. Made it direction-aware, mirroring `should_join_items`'s own pattern (`if prev.x <= curr.x { ... } else { ... }`).

**Strikeout as hard boundary (P2)**: Confirmed, and worth noting precisely what this closes: the render loop already excludes `<u>` on a struck item unconditionally, so an *overlapping* `<u>`/`<s>` span was never actually reachable in this version — but a struck fragment's flag could still skew the majority vote for its *non-struck* neighbors before this fix. Made strikeout a hard grouping boundary so a struck item's flag can't leak into a neighboring run's tally.

**Previous-item trailing punctuation (P2)**: Confirmed, the check was one-sided. Added the symmetric check on the previous item's own trailing character.

**Apostrophe breaks contractions (P2)**: Confirmed — good catch. Removed `'` from the closing-punctuation set; unlike `.`/`,`/etc. it can legitimately sit mid-word.

Added one regression test per finding (5 new), each verified to fail without its corresponding piece of the fix and pass with it. Full suite passing, clippy clean.

MADENIYOU:
Confirmed and fixed — good catch, and thanks for citing the exact \`should_join_items\` special cases to check against.

The run-grouping check was making its own independent judgment call using only geometry + punctuation, with no connection to \`should_join_items\` — the function the render loop's own \`needs_space_between\` call actually uses. Added it as a required condition for run continuation, so grouping can never disagree with the real spacing decision made in the same render path (CID-font one-word-per-operator gaps, the colon-label case, and anything else \`should_join_items\` already knows about).

New test \`underline_grouping_agrees_with_should_join_items_on_cid_fonts\`, reproducing the CID-font case cited in the finding (two distinct C2_-font words with a near-zero gap between them). Verified it fails without the fix (the two words wrongly merge into one underline run: \`<u>Alpha Beta</u>\` instead of \`<u>Alpha</u> Beta\`).

Full suite passing, clippy clean.

MADENIYOU:
Confirmed and fixed. \`needs_space_between\`'s hyphen exception is applied on top of \`should_join_items\`'s result, not inside it — \`should_join_items\` itself has no hyphen special-case. My previous fix mirrored only \`should_join_items\`, missing that outer exception.

New test \`underline_grouping_preserves_hyphen_no_space_exception\`, using the same CID-font shape as the previous finding's test (so \`should_join_items\` alone says "not joined") with a hyphenated split ("auto-" + "fix"). Verified it fails without the fix (splits into two runs at the hyphen: \`<u>auto-</u>fix\`) and passes with it.

Full suite passing, clippy clean.

PR #414, opened 2026-08-17
feat(markdown): opt-in inline-formula recovery as LaTeX from native math glyphs

abimaelmartell:
## Summary

Model-free formula extraction: rewrites native math glyph runs into `$...$` LaTeX during markdown conversion, using font evidence and item geometry — no recognition model, no downloads, no inference cost. Opt-in via `MarkdownOptions::formula_latex` / `pdf2md --formula-latex` (default off).

Supersedes the approach in #40/#42: detection is native and inline (no external layout-model bboxes), and emission goes into the markdown stream rather than a separate API. #42's unicode map is salvaged and extended (codepoint-variant siblings like MICRO SIGN → `\mu`, double-struck/script letters, ~60 more relations and arrows).

## How it works

- **Detection**: per-item evidence — *strong* (math font, or majority mapped math symbols, excluding prose punctuation like dot-leader ellipses and footnote daggers), *connective* (digits, brackets, relations, short italic identifiers, mixed identifier-punctuation like `F(q,r)=` — join but never anchor), *prose* (terminates). Detached script-fragment lines are stitched back to their base first: the line grouper's 3pt tolerance is smaller than TeX's superscript rise.
- **Reconstruction**: baseline+size medians classify sub/superscripts; adjacent same-kind script groups merge (`E_{4x4}`, never invalid `E_{4}_{x4}`); combining accents wrap the preceding glyph (`\hat{\alpha}`); font names select alphabets (MSBM → `\mathbb`, CMSY capitals → `\mathcal`, EUFM → `\mathfrak`); trailing sentence punctuation stays outside the delimiters.
- **Confidence gating**: stacked structure (fractions, matrices, bounded big operators), unbalanced delimiters, dangling accents, and unmapped glyphs all reject — the original Unicode text is kept. A wrong `$...$` is worse than none.

## Why opt-in

Measured on olmOCR-bench's render-equivalence math tests: converts 2.4% of arxiv_math from a hard zero (no local parser scores above zero there without LaTeX output), but LaTeXifying text currently costs fuzzy text-presence tests more than the math gains. Regression corpus with the flag on: all pass, with rewrites confined to math-heavy documents and all emitted LaTeX valid. Default stays off until net-positive; two known unlocks are queued:

1. Resolve BaseFont family names into extraction — `TextItem::font` is often an opaque resource name ("F2"), so font anchoring rarely fires today and detection leans on symbol anchoring.
2. Assemble display equations across full lines rather than intra-line runs (the render-equivalence bar requires whole expressions).

## Test plan

- [x] 33 unit tests in the new module: font profiles, run anchoring/rejection (dot leaders, footnote markers, prose), sub/superscript reconstruction, script stitching in both directions, accent wrapping, group merging, delimiter balancing, emphasis stripping
- [x] `cargo fmt` / `cargo clippy -- -D warnings` (both feature sets) / full suites (977 unit + 165 integration)
- [x] Default-off verified end to end: zero output change without the flag; regression corpus byte-identical

---
## Summary by cubic
Adds opt‑in inline LaTeX recovery for native math glyph runs during Markdown conversion. Previously output used raw Unicode only; with `MarkdownOptions.formula_latex` or `pdf2md --formula-latex` set, confident math runs emit `$...$` while prose stays unchanged. Default off keeps output identical.

- Hooks and flag: calls `formula::rewrite_math_runs` in both Markdown paths; adds `--formula-latex` and `MarkdownOptions::formula_latex`.
- Detection: anchors on math fonts or mapped symbols; digits/brackets/short italic identifiers can join runs; stitches detached script lines to their base; recognizes Latin Modern math families (`LMMathItalic`, `LMMathSymbols`, `LMMathExtension`, `LMMathOperators`).
- Reconstruction: maps Unicode to LaTeX, infers sub/superscripts from baseline/size, wraps combining accents, merges adjacent script groups, and places trailing sentence punctuation outside `$...$`.
- Confidence gating: rejects stacked structures, big operators with limits, unbalanced delimiters, unmapped glyphs, and risky LaTeX‑active ASCII; rewritten items clear emphasis flags to avoid splitting `$...$`.
- Rollout: no change by default. To enable, pass `--formula-latex` or set the option; downstream consumers should handle LaTeX in output when enabled.

<sup>Written for commit a4efd0aa9d3969ff11f1b32da96bfb050712201d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/pdf-inspector/pull/414?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/bin/pdf2md.rs
+++ b/src/bin/pdf2md.rs
@@ -406,6 +406,9 @@ fn main() {
             "  --compact           Collapse token-heavy source formatting such as dot leaders"
         );
         eprintln!("  --pages             Insert page break markers (<!-- Page N -->)");
+        eprintln!(
+            "  --formula-latex     Rewrite native math glyph runs into $...$ LaTeX (experimental)"
+        );
         eprintln!("  --select-pages N    Only process specified pages (e.g. 1,3,5-10)");
         eprintln!("  --password PW       Password for an encrypted PDF");
         eprintln!("  --detect-only       Only detect PDF type (no extraction)");
@@ -425,6 +428,7 @@ fn main() {
     let raw_output = args.iter().any(|a| a == "--raw");
     let compact_output = args.iter().any(|a| a == "--compact");
     let page_numbers = args.iter().any(|a| a == "--pages");
+    let formula_latex = args.iter().any(|a| a == "--formula-latex");
     let detect_only = args.iter().any(|a| a == "--detect-only");
     let analyze = args.iter().any(|a| a == "--analyze");
     let ocr_mode_argument = argument_value(&args, "--ocr").unwrap_or_else(|error| {
@@ -542,6 +546,7 @@ fn main() {
                 markdown.profile = pdf_inspector::MarkdownProfile::Compact;
             }
             markdown.include_page_numbers = page_numbers;
+            markdown.formula_latex = formula_latex;
             let mut pdf_options = OcrPdfOptions::new()
                 .render(RenderOptions::new().dpi(dpi))
                 .ocr(ocr)
@@ -617,6 +622,7 @@ fn main() {
         options.markdown.profile = pdf_inspector::MarkdownProfile::Compact;
     }
     options.markdown.include_page_numbers = page_numbers;
+    options.markdown.formula_latex = formula_latex;
     if let Some(pages) = page_filter {
         options.page_filter = Some(pages);
     }
--- a/src/formula/mod.rs
+++ b/src/formula/mod.rs
@@ -102,6 +102,9 @@ fn font_profile(font: &str) -> FontProfile {
         ("stmary", MathAlphabet::Plain),
         ("mathitalic", MathAlphabet::Plain),
         ("math-italic", MathAlphabet::Plain),
+        ("mathsymbols", MathAlphabet::Plain),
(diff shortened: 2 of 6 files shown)


abimaelmartell:
Measured with #415 merged in (font family names on items): font-evidence anchoring now fires on arXiv PDFs (LMMathItalic/MSBM/LMMathSymbols), and arxiv_math conversion rises from 2.4% to **6.4%** (188/2,927 render-equivalence tests, from a hard 0 before this PR). Remaining headroom is display-equation assembly across full lines — the render-equivalence bar requires whole expressions, and intra-line runs still fragment the left/right sides of large equations.

abimaelmartell:
Converting to draft. The inline reconstruction is measured and stable behind its flag, but the complete version of this feature depends on assembling display equations across lines, which may reshape parts of this module. Parking until that work is scheduled rather than carrying a dormant flag on main.

PR #421, opened 2026-08-18
Support dense CAD content streams

### LINKED ISSUE #307: Dense CAD/DWG PDF pages can return empty text when content streams exceed 1M operations
Summary: Some large native-text CAD/Revit-generated PDF pages contain more than 1,000,000 content-stream operations. pdf-inspector currently aborts extraction for the entire page once that guard is exceeded, which makes a text-based page appear to have no extractable text.

Observed behavior: A real structural drawing page with ~1.31M operations returned 0 positioned text items, despite containing selectable native PDF text. Another page with ~1.07M operations behaved the same way.

Expected behavior: 
Either process the page safely with a higher/configurable operation limit, or
return an explicit “operation limit exceeded” diagnostic rather than an indistinguishable empty-text result.

Local fix tested: Raising the finite guard from 1,000,000 to 2,000,000 allowed both pages to extract successfully while retaining protection against pathological streams. Synthetic regression tests also verify that a >2M page is still safely skipped.

Results:
~1.31M-op page: 0 → 1,351 positioned items
~1.07M-op page: 0 → 614 positioned items

yzxcj797:
## Summary

- Raise the page content-stream safety budget from 1,000,000 to 2,000,000 operators.
- Preserve the hard fail-closed boundary for pathological streams above the new limit.
- Propagate a machine-readable `content_operation_limit` OCR reason when a page is skipped by that budget, instead of making it indistinguishable from a genuinely empty page.
- Add regression coverage for the reported 1.1M-operator CAD shape, text survival below the limit, fail-closed behavior above it, and public per-page diagnostics.

Fixes #307.

## Rationale

The existing bounded decoder protects against unbounded operation allocation, but the one-million threshold also rejects real CAD/Revit pages containing native text. The issue reports valid pages at approximately 1.07M and 1.31M operators. A two-million ceiling covers those documents while retaining a finite allocation guard.

The new reason is also deliberately distinct from `no_text`: callers can tell that extraction was stopped by the safety budget and choose an appropriate fallback.

## Testing

- `cargo test`
  - 954 library tests pass
  - 163 integration tests pass
  - 3 CLI tests pass
  - 2 documentation tests pass
- `cargo clippy -- -D warnings` — passes
- `cargo fmt --all -- --check` — passes
- `git diff --check` — passes

The optional OCR runtime suites were not run locally; this change does not alter OCR runtime code.

---
## Summary by cubic
Supports dense CAD/Revit PDFs by doubling the content-stream operator budget and emitting a dedicated OCR reason when the limit is hit. Previously, pages over 1,000,000 operators looked like empty text and could imply encoding issues; now the cap is 2,000,000 and over-cap pages are skipped with reason "content_operation_limit" without flagging encoding corruption.

- Adds OCR reason "content_operation_limit" and surfaces it in per-page diagnostics and `ocrReasonsByPage`; `extractPagesMarkdown` sets the page’s `ocr_reason` accordingly.
- Tracks pages skipped by the operator budget and includes them in `pages_needing_ocr`; does not set `has_encoding_issues` when this is the only reason.
- Introduces `extract_page_text_items_with_operation_limit`; makes `content_exceeds_operation_limit` pub(crate); adds regression tests for under/over-cap behavior.
- Migration: if you surface OCR reasons, accept "content_operation_limit"; no API shape changes to `@firecrawl/pdf-inspector`.
- Addresses #307.

<sup>Written for commit 4229cf6d512d823b645ed6fecaadeb69e9008bd9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/pdf-inspector/pull/421?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/napi/README.md
+++ b/napi/README.md
@@ -94,7 +94,7 @@ console.log(result.confidence)     // 0.875
 
 Extract text within bounding-box regions from a PDF. Designed for hybrid OCR pipelines where a layout model detects regions in rendered page images, and this function extracts text from the PDF structure for text-based pages — skipping GPU OCR.
 
-Each region result includes a `needsOcr` flag that signals unreliable extraction (empty text, GID-encoded fonts, garbage text, encoding issues). Known causes are exposed in `ocrReason`; for example, a suspected garbled text layer reports `"suspected_garbled_text"`, and a page skipped by the content-stream safety budget reports `"content_operation_limit"`.
+Each region result includes a `needsOcr` flag that signals unreliable extraction (empty text, GID-encoded fonts, garbage text, encoding issues). Known region causes are exposed in `ocrReason`; for example, a suspected garbled text layer reports `"suspected_garbled_text"`. A page skipped by the content-stream safety budget reports `"content_operation_limit"` in the page-level `ocrReasonsByPage` result.
 
 ```typescript
 import { extractTextInRegions } from '@firecrawl/pdf-inspector'
--- a/napi/README.md
+++ b/napi/README.md
@@ -94,7 +94,7 @@ console.log(result.confidence)     // 0.875
 
 Extract text within bounding-box regions from a PDF. Designed for hybrid OCR pipelines where a layout model detects regions in rendered page images, and this function extracts text from the PDF structure for text-based pages — skipping GPU OCR.
 
-Each region result includes a `needsOcr` flag that signals unreliable extraction (empty text, GID-encoded fonts, garbage text, encoding issues). When the cause is a suspected garbled text layer, `ocrReason` is set to `"suspected_garbled_text"`.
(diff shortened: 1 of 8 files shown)


yzxcj797:
Addressed in 4229cf6:

- `process_pdf_mem` now includes budget-skipped pages in `pages_needing_ocr`.
- `content_operation_limit` no longer by itself sets `has_encoding_issues`; actual text-quality or encoding signals still do.
- Added a process-level regression covering both invariants.
- Corrected the Node documentation: the budget reason is page-level (`ocrReasonsByPage`), not a region-level `ocrReason`.

Validation: full `cargo test`, `cargo clippy -- -D warnings`, `cargo fmt --all -- --check`, and `git diff --check` pass.

PR #424, opened 2026-08-18
Keep adjacent numeric table columns separate

### LINKED ISSUE #419: Distinct numeric value columns merged into one cell for one row per PDF (customs FDI statistical tables)
# Bug report: distinct numeric value columns merged in heuristic table extraction

## Summary

Two adjacent numeric value columns (`Reporting month.Value (USD)` and `Year to date.Volume`) are merged into a single cell for exactly one row per PDF, on Vietnamese Customs FDI statistical reports (`Table: 17B/TCHQ`). Reproduced on **pdf-inspector v1.15.0** (latest release, 2026-08-17) and on `master` (grid.rs unchanged since the merge pass was introduced).

## Environment

- pdf-inspector v1.15.0 (pip `pdf-inspector`), Python 3.13
- `extract_pages_markdown()` on a text-based 2-page PDF (78 KB, `classify_pdf → text_based, confidence 1.00`)

## Repro

Fixture: customs FDI preliminary report — July 2026 import, row 6 "Coal" (also reproduced on June 2026 import PDF; both attached). Both are real production PDFs from the Vietnam General Department of Customs.

```python
from pdf_inspector import extract_pages_markdown

r = extract_pages_markdown("july_import.pdf")
md = "\n".join(p.markdown for p in r.pages)
for line in md.splitlines():
    if "|6 Coal|" in line:
        print(line)
```

Actual (bug):

```
|6 Coal|Ton|1,709,851|246,515,548 11,734,697||1,555,338,055|
```

Expected: `rep_val` and `ytd_vol` as separate cells:

```
|6 Coal|Ton|1,709,851|246,515,548|11,734,697|1,555,338,055|
```

Every other row in the same table extracts correctly (34 rows total; only row 6 merges). Export-direction PDFs (different table layout) extract cleanly — 0 merged rows. The same Coal-row defect reproduces on June 2026 import: `|6 Coal|Ton|1,661,481|236,480,212 10,024,724||1,302,660,618|`.

## Root cause analysis

PDF word geometry (from `pdfplumber`, x0/x1 coordinates in pt) for the Coal row vs a clean row:

```
Wheats row value gaps: [22.5, 13.6, 20.6]  min 13.6 pt
Coal   row value gaps: [17.0,  8.1, 12.3]  min  8.1 pt   <-- abnormal
```

The Coal row's `rep_val → ytd_vol` gap is **8.1 pt** (every other row ≥ 12.3 pt). This is deterministic in the source PDF (same 8.1 pt signature in both June and July; the Coal commodity name is short, shifting the value-column alignment).

In `src/tables/grid.rs`, `merge_numeric_adjacent_clusters`:

```rust
let merge_dist = threshold * 1.5;   // default threshold 25pt -> 37.5pt
// merge if dense.numeric_frac > 0.50 && sparse.count <= dense.count / 2 && sparse.count <= 5
```

Both adjacent clusters are predominantly numeric; the cluster centers land within `merge_dist` (37.5 pt) because the Coal row's 8.1 pt gap pulls the centers together, so the merge pass collapses the two value columns for this row. For all other rows the gap (≥12.3 pt) keeps the centers far enough apart to avoid the merge — but only just (12.3 pt is still well under 37.5 pt; the algorithm happens to survive on center distribution, not on a hard gap check).

## Suggested fix

The numeric merge pass exists to fix header-vs-data splits (`"header text"` next to a dense numeric column). Merging **two predominantly numeric clusters** into one is never the intended behavior. Propose guarding `merge_numeric_adjacent_clusters` so it does not merge when **both** clusters are predominantly numeric:

```rust
let should_merge =
    dense.numeric_frac > 0.50
    && sparse.count <= dense.count / 2
    && sparse.count <= 5
    && !(info_a.numeric_frac > 0.50 && info_b.numeric_frac > 0.50); // new guard
```

This preserves the header-fix behavior while preventing numeric-column collapse. Alternative (weaker): require `dist > some_min_gap` when both numeric.

## Fixtures

- `july_import.pdf` (78,419 bytes) — July 2026, import, reproduces
- `2026_06_import_preliminary.pdf` (78,347 bytes) — June 2026, import, reproduces
- Control (clean): `2026_06_export_preliminary.pdf` — export direction, no merge

yzxcj797:
## Summary

- Keep two predominantly numeric column clusters separate instead of collapsing the sparse numeric column into the dense one.
- Preserve the existing header repair path: a sparse predominantly non-numeric cluster can still merge with adjacent numeric data.
- Split a two-value financial item only when its physical span crosses a ruled column boundary. The existing heuristic splitter remains limited to wider items with three or more values, so this does not broadly relax heuristic-table splitting.
- Handle empty/degenerate ruled grids safely.

Fixes #419.

## Validation on the attached customs PDFs

Before:

```text
June: |6 Coal|Ton|1,661,481|236,480,212 10,024,724||1,302,660,618|
July: |6 Coal|Ton|1,709,851|246,515,548 11,734,697||1,555,338,055|
```

After:

```text
June: |6 Coal|Ton|1,661,481|236,480,212|10,024,724|1,302,660,618|
July: |6 Coal|Ton|1,709,851|246,515,548|11,734,697|1,555,338,055|
```

## Tests

- `cargo test`
  - 958 library tests pass
  - 162 integration tests pass
  - 3 CLI tests pass
  - 2 documentation tests pass
- `cargo clippy -- -D warnings` — passes
- `cargo fmt --all -- --check` — passes
- `git diff --check` — passes

---
## Summary by cubic
Keep adjacent numeric table columns separate and split two-value financial items only when they cross a ruled column boundary. Preserves header-to-data merging and inferred-grid behavior. Fixes #419.

- Adjacent-column merge: merge a sparse cluster into an adjacent dense cluster only when the sparse cluster is predominantly non-numeric; two predominantly numeric clusters now always remain separate. Classification uses the actual members of each cluster to avoid jitter reclassification.
- Ruled-boundary split: allow splitting a two-value numeric item only in ruled grids and only for `Text` items with positive font size; keep the heuristic splitter unchanged (requires 3+ values). Preserve slot alignment (heuristic: centered; ruled: slot start), and fall back to the original item if any split fragment cannot be placed.
- Grid assignment: introduce `assign_items_to_ruled_grid` for ruled tables; inferred grids continue to use `assign_items_to_grid` so two-value items stay intact. Deduplicate item indices and ignore zero-width items.
- Tests: add coverage to keep numeric columns separate, still merge sparse headers with numeric data, and split only in ruled grids.

<sup>Written for commit 4d90870f24771d337d28b9e9df9635ecd684237d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/pdf-inspector/pull/424?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/tables/detect_lines.rs
+++ b/src/tables/detect_lines.rs
@@ -9,7 +9,7 @@ use std::collections::{HashMap, HashSet};
 use crate::tables::Table;
 use crate::types::{PdfLine, PdfRect, TextItem};
 
-use super::detect_rects::{assign_items_to_grid, snap_edges};
+use super::detect_rects::{assign_items_to_ruled_grid, snap_edges};
 
 const RULE_Y_TOLERANCE: f32 = 2.0;
 const RULE_JOIN_GAP: f32 = 6.0;
@@ -956,7 +956,8 @@ fn build_open_edge_grid_table_for_rules(
         return None;
     }
 
-    let (body_cells, mut item_indices) = assign_items_to_grid(items, &col_edges, &row_edges, page);
+    let (body_cells, mut item_indices) =
+        assign_items_to_ruled_grid(items, &col_edges, &row_edges, page);
     let column_count = col_edges.len() - 1;
     let occupied_body_rows = body_cells
         .iter()
@@ -1772,7 +1773,8 @@ fn detect_tables_from_lines_inner(
     );
 
     // Assign items to grid
-    let (cells, item_indices) = assign_items_to_grid(items, &col_edges, &row_edges_desc, page);
+    let (cells, item_indices) =
+        assign_items_to_ruled_grid(items, &col_edges, &row_edges_desc, page);
 
     // Require at least 2 non-empty rows
     let non_empty_rows = cells
--- a/src/tables/detect_rects.rs
+++ b/src/tables/detect_rects.rs
@@ -6,6 +6,7 @@ use log::debug;
 
 use crate::types::{PdfRect, TextItem};
 
+use super::financial::try_split_financial_item_across_rule;
 use super::Table;
 
 const DOMINANT_PAGE_BACKGROUND_MIN_REPETITIONS: usize = 8;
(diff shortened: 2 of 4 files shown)


yzxcj797:
Addressed in 4d90870:

- Cluster numeric fractions now consume the exact item X values belonging to each cluster, so adjacent-cluster jitter cannot reclassify a cluster through an overlapping radius window.
- Added the reported mixed-jitter shape as a regression while retaining the header-vs-data merge case.
- Two-value splitting is now limited to ruled grids via a dedicated assignment entry point; inferred row-stripe and merged-cluster grids keep the original consolidated item.
- The splitter accepts only positive-font-size text items, rejecting form fields.
- The heuristic splitter keeps its original centered sub-item placement; only the ruled-boundary splitter uses left-aligned slots.
- If any split fragment cannot be assigned to a column, the original item is kept together instead of marking it consumed and losing the unplaced value.
- Added regression coverage for inferred grids, form fields, and splitter-specific alignment.

Validation: full `cargo test`, `cargo clippy -- -D warnings`, `cargo fmt --all -- --check`, and `git diff --check` pass. Both attached customs PDFs still split the Coal row correctly.

Issue #214, opened 2026-08-02
Feature: language-aware sanity signal for text layers that decode to the wrong characters


al-ashalash:
## Problem

The most damaging failure mode we see in practice is not a missing text layer — it is a text layer that is **present and decodes cleanly to valid Unicode, but to the wrong letters**, because the embedded `ToUnicode` CMap is wrong rather than absent.

The extracted text passes every structural check: no replacement characters, valid script ranges, plausible word lengths. `hasEncodingIssues` returned `false` on **all** of our affected samples. Callers therefore skip OCR and receive convincing garbage — worse than a missing text layer, because nothing downstream can tell.

This is common in Arabic government and court documents that were scanned and OCRed by other software before being distributed as PDFs. The page reads `المدعى عليه` and the text layer yields `الددعى عليهي` — a systematic single-letter substitution that no structural check can catch.

## Relationship to #122 / PR #195 and #118

#122 (and PR #195) expose whether a font *has* a `ToUnicode` map, and #195 states plainly that "broken-but-present ToUnicode stays `false`". #118 covered a constant per-range character shift with `ToUnicode` present, and neither `is_garbage_text` nor `detect_encoding_issues` caught it.

That gap — **present but wrong** — is exactly what this issue is about, and it cannot be closed structurally. Something has to look at the text and ask whether it reads like the language it claims to be.

## Suggested approach: score the text against a small common-word list

We ship this check in an Arabic document tool and calibrated it on real files. One pass over the text against a ~60-word set per language, no model, negligible cost.

Calibration on our corpus (document labels kept generic on purpose):

| document type | ratio | true state |
|---|---|---|
| peer-reviewed article | 0.172 | sound |
| peer-reviewed article | 0.162 | sound |
| university course notes | 0.156 | sound |
| court document | 0.146 | corrupt |
| journal article | 0.120 | corrupt |
| journal article | 0.097 | corrupt |
| court document | 0.050 | corrupt |

A threshold of 0.15 separated them cleanly.

## Implementation details that matter — corrections to my own recipe above

I originally wrote this as "tokenize → count → divide". Reviewing our own implementation, that summary is incomplete in four ways, and anyone implementing it literally would ship something that misfires. Correcting it here rather than letting it be discovered the hard way:

[...]

Issue #212, opened 2026-08-02
Arabic (RTL) text is returned in visual order, not logical order


al-ashalash:
## Summary

For right-to-left scripts, `extractText` returns each line's characters in **visual** order rather than logical order, so the output is reversed and unusable as-is.

This matters more than a formatting nit: the library correctly identifies these files as text-based and tells the caller to skip OCR, but the text it hands back cannot be used. A silently wrong result is worse than a "needs OCR" verdict. README.md, `napi/README.md` and `docs/rust-api.md` all currently advertise "RTL text support", so callers have no reason to check.

Tested on v1.11.2, Windows x64 (`@firecrawl/pdf-inspector-win32-x64-msvc`).

## Reproduction

No sample file needed — you can generate one:

1. Save as `ar.html`:

```html
<html dir="rtl"><head><meta charset="utf-8"></head><body>
<p>السلام عليكم ورحمة الله. هذه فقرة اختبارية مكتوبة باللغة العربية.</p>
<p>الطلاب يدرسون خلال الفصل الدراسي، والكلام الواضح خير من الإطالة.</p>
<p>درجة الحرارة اليوم 25 مئوية والرطوبة 60 بالمئة.</p>
</body></html>
```

2. `msedge --headless --print-to-pdf=ar.pdf ar.html` (Chrome works identically)
3. `console.log(extractText(fs.readFileSync('ar.pdf')))`

## Expected

```
السلام عليكم ورحمة الله. هذه فقرة اختبارية مكتوبة باللغة العربية.
```

## Actual

```
.ﺔﻴﺑﺮﻌﻟا ﺔﻐﻠﻟبا ﺔﺑﻮﺘﻜﻣ ﺔﻳرﺎﺒﺘﺧا ةﺮﻘﻓ ﻩﺬﻫ .ﷲ ﺔﲪرو ﻢﻜﻴﻠﻋ مﻼﺴﻟا
```

Characters run in visual order, and Arabic presentation forms (U+FB50–U+FEFF) are emitted as-is instead of their standard code points.

## Where this happens in the source

I went looking after filing, and `src/text_utils.rs` explains all three symptoms. Quoting what I read there (please correct me if I've misread the flow):

**a) Reversal is gated on presentation forms being present.**

```rust
let had_presentation_forms = text.chars().any(is_arabic_presentation_form);
…
if had_presentation_forms {
    result = reverse_visual_arabic(&result);
}
```

[...]

PR #447, opened 2026-08-21
test(rtl): add generated Arabic fixtures guarding visual-order extraction

### LINKED ISSUE #218: Add RTL/Arabic regression fixtures to CI — generated at test time, nothing to host
## Problem

There is no Arabic or RTL document in the test matrix, and `pdf-evals` snapshots can't easily cover it — real Arabic documents are mostly copyrighted or private, so there is nothing convenient to host.

That is a problem right now: #212 and PR #207 are both in flight against `text_utils.rs`, and whichever lands, nothing will catch it silently regressing later. RTL bugs are especially prone to that, because the output *looks* like text to a reviewer who doesn't read the script — the failure is invisible in a diff.

## Proposal: generate the fixtures instead of hosting them

Print HTML to PDF with headless Chrome at test setup. No sample files in the repo, no licensing question, byte-stable across runs, and works in Linux CI:

```
chrome --headless --disable-gpu --no-pdf-header-footer --print-to-pdf=ar.pdf ar.html
```

We use exactly this in our own test suite and the same trick is the reproduction in #212, so it is already proven to produce the failing behaviour.

## Three assertions worth copying

These are the ones that earned their keep in our suite:

**1. Digit groups must survive reordering intact.** Assert on a string like `126,248.34` embedded in an Arabic line — it catches digit-run reversal, thousands separators being split, and the decimal point migrating, all in one assertion. This is a live defect: see #212 (`25` → `52`) and #200.

**2. Ligature round-trip.** A line containing `لا` sequences (`خلال`, `السلام`, `الكلام`) must come back character-identical. Multi-character `ToUnicode` expansions are where naive reversal breaks, and it breaks quietly — the text still reads as Arabic, just wrong.

**3. A negative control.** This is the one people skip. Run the same input with the RTL path disabled and assert the output is **wrong**. If a test passes both with and without the fix, it isn't testing the fix — it's testing that the pipeline runs. We caught one of our own tests doing exactly that.

## What I can contribute

I can send a PR with:

- a fixture generator (HTML → PDF via headless Chrome, skipping gracefully when Chrome is absent),
- the three assertions above against `extractText` and `extractTextWithPositions`,
- a synthetic corruption generator: apply a fixed letter-substitution map to sound Arabic text to fabricate a "wrong ToUnicode" text layer, so #214-style detection can be tested without hunting for a genuinely corrupt file.

Before I do — two questions so it lands in the right shape:

1. Do you want this in this repo's `cargo test`, or in the `pdf-evals` sibling repo alongside the snapshot corpus?
2. Is a headless-Chrome dependency acceptable in CI, or would you rather I emit the PDFs from a checked-in byte template so the tests have no external dependency at all? The second is more work but keeps CI hermetic — happy either way.

Also relevant: `AGENTS.md` notes that character-level diff alone misclassifies structural improvements as regressions and that `bench.py score` is the tie-breaker. For RTL that distinction is sharper than usual — a fix that corrects word order will look like a large diff and score as a regression under a naive comparison. Worth deciding up front which harness these fixtures answer to.

mathurshubham:
Implements #218: Arabic/RTL regression coverage with fixtures generated at test time — no hosted sample files, no licensing question, and no headless-Chrome dependency in CI.

## Why now

#440 just landed geometric visual-order reversal in `fix_visual_order_rtl`. As #218 points out, RTL bugs are uniquely prone to silent regression: the output *looks* like text to a reviewer who doesn't read the script, so a revert or refactor that breaks it is invisible in a diff. These tests are the guard.

## How the fixtures are generated

Pure Rust, in-test (`tests/rtl_generated_tests.rs`, no production source touched). `make_rtl_pdf(lines, storage)` hand-builds a single-page PDF with a byte-accurate xref — the same pattern as the repo's existing `make_text_pdf` helpers and the same shape as the Hebrew fixtures that landed with #440:

- A `/TrueType` font with a `/ToUnicode` CMap maps single-byte codes to Unicode, so no real Arabic glyph program is embedded.
- Each token is one `Tj` with an absolute `Tm`. **Visual-order** storage reverses each Arabic run's characters and paints left-to-right (increasing x — the case #440 must reverse). **Logical-order** storage stores reading order and paints right-to-left (the case #440 must leave alone). Both the geometric vote and `restore_embedded_ltr_runs` are exercised.
- Every test word carries a lam-alef pair emitted as a single glyph code mapping to two code points (U+0644 U+0627), covering the multi-character ToUnicode expansion where naive reversal breaks.

## Assertions (from #218)

1. **Digit groups survive reordering intact** — `126,248.34` inside an Arabic line must come out unreversed: `arabic_visual_order_keeps_digit_group_intact`, `arabic_logical_order_keeps_digit_group_intact`, `arabic_digit_group_intact_in_positioned_items`.
2. **Multi-char ToUnicode ligature round-trip** — `arabic_visual_order_ligature_round_trip`, `arabic_logical_order_ligature_round_trip`.
3. **Negative control** — the character-reversed corruptions (`مالسلا`, `مالكلا`) must be absent from output: `visual_order_negative_control_wrong_output_is_rejected`.

Plus `visual_and_logical_storage_agree`: both storage conventions extract to identical reading order. Both `process_pdf_mem` (markdown) and `extract_text_with_positions_mem` (positioned items) are covered.

## Revert check

With `fix_visual_order_rtl` temporarily neutralized (early return), the 4 visual-order tests fail with the expected reversed corruption (e.g. `## لالخ مالسلا مالكلا`) while the 3 logical-order/positioned tests still pass — exactly the split the guard should produce. All 7 pass on current `main`.

## Verification

Full `cargo test` green (1015 lib + 165 integration + 7 new rtl_generated + 3 bin + 2 doc-tests, 0 failures); `cargo fmt` and `cargo clippy -- -D warnings` clean.

Out of scope, per the issue discussion: the synthetic wrong-ToUnicode corruption generator (that targets #214's detection feature, not this reversal guard) and Arabic-Indic digit variants (the ASCII digit-group case from the issue is asserted; #440's logic covers both scripts).

Closes #218.

---
## Summary by cubic
Adds generated Arabic/RTL regression tests that synthesize PDFs at test time to guard visual-order text extraction. Prevents silent regressions in `fix_visual_order_rtl` (#440) without hosting sample files or adding headless browser deps. Closes #218.

- **Review notes**
  - Test-only change: new `tests/rtl_generated_tests.rs` builds single-page PDFs in Rust with a `/ToUnicode` CMap; exercises both visual and logical storage and `restore_embedded_ltr_runs`.
  - Assertions per #218: digit groups (`126,248.34`) remain unreversed inside RTL lines; lam‑alef (`لا`) multi-character `ToUnicode` expansion round-trips; negative control rejects raw reversed forms; both `process_pdf_mem` and `extract_text_with_positions_mem` agree; visual and logical storage extract to identical reading order.
  - Guard verified: disabling `fix_visual_order_rtl` makes visual-order tests fail while logical/positioned tests pass.

<sup>Written for commit f856f3f765f03bf8f709d8fa43cc901a57290456. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/pdf-inspector/pull/447?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/tests/rtl_generated_tests.rs
+++ b/tests/rtl_generated_tests.rs
@@ -1,0 +1,416 @@
+//! Generated Arabic / RTL regression fixtures.
+//!
+//! There are no real Arabic documents in the test matrix — genuine ones are
+//! almost all copyrighted or private, so nothing can be hosted (issue #218).
+//! Instead of shipping sample files (or standing up a headless-Chrome
+//! dependency in CI), these tests synthesize the PDFs in pure Rust at test
+//! time, following the same hand-rolled `%PDF` + `ToUnicode` CMap approach the
+//! repo already uses for the checked-in Hebrew fixtures that landed with
+//! PR #440 ("reverse visual-order RTL text using glyph geometry").
+//!
+//! They are the regression guard for #440: base-alphabet RTL text stored in
+//! *visual* (screen left-to-right) order must be reversed back to logical
+//! reading order, while text stored in *logical* order must be left alone. A
+//! codepoint-only trigger would corrupt one convention or the other, so both
+//! storage conventions are generated and both must extract to the same logical
+//! text.
+//!
+//! Assertions mirror the three that the issue calls out:
+//!   1. digit groups (`126,248.34`) embedded in an Arabic line survive intact
+//!      and unreversed,
+//!   2. `لا` (lam-alef) sequences delivered as a single multi-character
+//!      `ToUnicode` expansion round-trip character-identical,
+//!   3. a negative control — the visual-order fixture must NOT come out in the
+//!      raw stored (reversed) order, so a test that passes with #440 reverted
+//!      would be caught.
+
+use pdf_inspector::{extract_text_with_positions_mem, process_pdf_mem};
+
+// ---------------------------------------------------------------------------
+// Test vocabulary
+// ---------------------------------------------------------------------------
+//
+// All words are base-alphabet Arabic (no presentation forms), so they take the
+// #440 geometry-reversal path rather than the older NFKC/presentation-form path
+// in `expand_ligatures`. Each of these words contains a `لا` (lam-alef) pair,
+// which the generator always emits as a single glyph code mapped to the two
+// code points U+0644 U+0627 — exercising the multi-character `ToUnicode`
+// expansion that naive reversal breaks.
+
+const SALAM: &str = "السلام"; // ا ل س ل ا م   (contains لا)
+const KALAM: &str = "الكلام"; // ا ل ك ل ا م   (contains لا)
+const KHILAL: &str = "خلال"; //  خ ل ا ل       (contains لا)
+const NUMBER: &str = "126,248.34"; // ASCII digit group with separators
+
+#[derive(Clone, Copy, PartialEq)]
+enum Storage {
+    /// Each show op stores its run's characters in screen (left-to-right)
+    /// order — reversed relative to reading order — and ops walk the line
+    /// left-to-right. This is how producers of *visible shaped* RTL text emit.
+    Visual,
+    /// Each show op stores one run in reading order and successive ops are
+    /// positioned right-to-left across the line (the OCR-text-layer
+    /// convention).
+    Logical,
+}
+
+fn is_arabic(s: &str) -> bool {
+    s.chars().any(|c| ('\u{0600}'..='\u{06FF}').contains(&c))
+}
+
+/// Split a string into glyph "units": a `لا` (U+0644 U+0627) pair becomes one
+/// unit (a single font glyph with a two-code-point `ToUnicode` mapping), every
+/// other character is its own unit.
+fn glyph_units(s: &str) -> Vec<String> {
+    let chars: Vec<char> = s.chars().collect();
+    let mut units = Vec::new();
+    let mut i = 0;
+    while i < chars.len() {
+        if chars[i] == '\u{0644}' && i + 1 < chars.len() && chars[i + 1] == '\u{0627}' {
+            units.push("\u{0644}\u{0627}".to_string());
+            i += 2;
+        } else {
+            units.push(chars[i].to_string());
+            i += 1;
+        }
+    }
+    units
+}
+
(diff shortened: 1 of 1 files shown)


al-ashalash:
Thanks for building this — the negative control in particular is the right instrument, and generating the fixtures in Rust removes the CI dependency I asked about in #218. Two notes, and I think the first is load-bearing.

**The suite does not cover the path the issue came from.** The header says so plainly:

> All words are base-alphabet Arabic (no presentation forms), so they take the #440 geometry-reversal path rather than the older NFKC/presentation-form path in `expand_ligatures`.

#218 was filed off the reproduction in #212, and that reproduction is a browser print-to-PDF, which embeds *shaped* presentation forms (U+FB50–U+FEFF). As it stands, the RTL regression suite guards the path #440 added and leaves untested the path the original bug lives on.

That matters more than a coverage gap, because I measured the published build today and posted the numbers in #212: `@firecrawl/pdf-inspector` 1.11.2 and 1.17.0 produce byte-identical output on that reproduction — still visual order. A green "RTL generated fixtures" suite landing next to an unfixed path is how a defect becomes invisible.

**Suggested addition, small.** A third axis on the generator — shaping — so `Storage::Visual` can be emitted either as base letters or as presentation forms, with the same `ToUnicode` expansion. The lam-alef case you already build is the natural carrier: U+FEFB is exactly the shaped form of the pair you map to U+0644 U+0627. Three tests would cover it.

If you would rather not grow the PR, the alternative that still closes the hole is one `#[ignore]`d test holding the expected reading-order output for the shaped case, with a comment pointing at #212. A named known-failing test is a far better record than silence, and it turns green the day the gate is widened.

Happy to supply the expected strings for whichever shape you pick.

PR #449, opened 2026-08-21
feat(extractor): expose the raw font resource tag on TextItem

abimaelmartell:
## Summary

1.16.0 replaced `TextItem::font`'s resource tag with the `/BaseFont` family name (#415). That is the right default, but it erased the only signal distinguishing two font programs that share a family name — consumers that partition items by font program (style grouping, feature parity with pre-1.16.0 extractions) lost their key when same-family subsets merged.

This adds a `font_tag` field carrying exactly what `font` carried before 1.16.0:

- the raw per-page resource tag (`F2`, `T22`), stamped by both content-stream parsers from the same resource name
- first run's tag on merged items; propagated through derived items (numeric splits, heuristic-table splits)
- the tag active at `EMC` for ActualText spans, matching the sibling `font`/`is_bold`/`is_italic`/`font_size` fields, which all sample the same state at the same point — legacy parity is the contract, so any change to that sampling must move all five fields together
- empty for items not born from a content-stream show operator (images, links, form fields, OCR)

Exposed through `--items-json`, the Node binding (`fontTag`), and the Python binding (`font_tag` + `.pyi` stub). The wasm binding doesn't surface positioned items, so no change there.

## Compatibility

- npm and PyPI surfaces are strictly additive; markdown output is byte-identical.
- For the Rust crate, the new pub field is source-breaking for consumers that construct `TextItem` literally or destructure it exhaustively (field readers are unaffected). `TextItem` is an output type, so literal construction outside tests is expected to be rare, but the release that ships this should be versioned with that caveat in mind.

## Notes

- Tests: stamping parity pinned in both parsers (`items_carry_family_name_and_resource_tag`, `form_items_carry_family_name_and_resource_tag`).


abimaelmartell:
@cubic-dev-ai

PR #292, opened 2026-08-07
feat(go): add pure Go bindings (CGO=0) via WebAssembly and wazero

Raina-Hardik:
## Summary

This PR adds pure Go (\CGO_ENABLED=0\) bindings for \pdf-inspector\ by embedding a release WebAssembly (\wasm32-wasip1\) module and driving it with [\wazero\](https://github.com/tetratelabs/wazero), a zero-dependency pure Go WebAssembly engine.

### Features
- **Zero CGO (\CGO_ENABLED=0\)**: No C toolchain or cross-compilation headers required for Go applications.
- **Embedded WASM Binary**: The \pdf_inspector.wasm\ binary is compiled and bundled into the Go module via \//go:embed\.
- **Full API Coverage**:
  - \pdfinspector.ProcessPdf(pdfBytes, opts)\: Complete PDF text, layout, and Markdown extraction.
  - \pdfinspector.DetectPdf(pdfBytes, password)\: Fast metadata analysis & OCR necessity checks.
  - \pdfinspector.ClassifyPdf(pdfBytes)\: Fast classification (\TextBased\, \Scanned\, \ImageBased\, \Mixed\).
  - \pdfinspector.ExtractText(pdfBytes)\: Raw plain-text extraction.
  - \pdfinspector.ExtractPagesMarkdown(pdfBytes, pages)\: Page-level Markdown extraction.
  - \pdfinspector.Version()\: WASM library version lookup.
- **Safe FFI & Memory Management**: WASM FFI entry points (\src/ffi.rs\) use \std::panic::catch_unwind\ to report structured errors without panicking host runtime execution.

### Changes Included
- \src/ffi.rs\: Rust FFI layer with standard allocation, deallocation, and JSON serialization routines.
- \Cargo.toml\ / \src/lib.rs\: Exposed FFI module for target \wasm32-wasip1\.
- \go.mod\ / \go.sum\: Standard Go module setup importing \wazero v1.8.2\.
- \pdfinspector.go\: Public Go API and WASM module runner.
- \pdfinspector_test.go\: Go unit test suite verifying extraction against test fixtures.

### Testing
Ran unit tests with \CGO_ENABLED=0\:
\go test -v .\ -> PASS (All 7 unit tests passing)

---
[...]

Issue #267, opened 2026-08-05
Expose configurable confidence threshold for OCR routing and document threshold behavior


yulinlina:
## Description

`pdf-inspector` reports a confidence score and per-page OCR routing, but it is not clear whether callers can adjust the threshold used to decide when a page or document should be routed to OCR. A small configuration option would make the routing behavior easier to tune for different workloads.

## Motivation

In production, the optimal threshold depends on OCR cost, latency, and acceptable extraction quality. Some users may prefer aggressive OCR routing for high recall, while others may prefer local extraction unless confidence is very low. Without an exposed threshold or documentation, users may need to re-implement routing logic or guess the library's default behavior.

## Proposed Solution

- Add an optional configuration field such as `ocr_confidence_threshold` to the classification API in Rust.
- Keep the current default behavior unchanged for backward compatibility.
- Expose the option in Python, Node.js, and WASM bindings where applicable.
- Document the meaning of confidence scores, per-page routing decisions, and how the threshold affects classification.
- Include short examples showing how to route pages based on the returned metadata and configured threshold.

Example Rust API idea:

```rust
let options = ClassificationOptions {
    ocr_confidence_threshold: 0.65,
    ..Default::default()
};
let result = inspect_pdf_with_options(path, options)?;
```

## Alternatives Considered

- Leave the threshold fixed and document only how to interpret the confidence score.
- Require callers to ignore the built-in routing decision and implement custom routing from the confidence score themselves.
- Provide a separate helper function for thresholding instead of adding configuration to the core API.

---
🤖 *Suggested by [GitHub Trend-Hacking Pipeline](https://github.com/yulinlina/start_rise)*

nirupam-06:
Hi! I'm interested in working on this issue. Could you please assign it to me? Thanks!

yulinlina:
Sure, please take this on. I may not be able to assign it from my side, so opening a draft PR or posting a short implementation note here would be a good way to claim it.

For the implementation, I’d suggest keeping the API additive and backward-compatible:

```rust
#[derive(Clone, Debug, Default)]
pub struct OcrRoutingConfig {
    /// Confidence threshold in [0.0, 1.0].
    /// Pages with confidence below this value may be routed to OCR.
    /// If `None`, preserve the current built-in default.
    pub confidence_threshold: Option<f32>,
}
```

A few important implementation points:

- Validate the threshold: reject values outside `0.0..=1.0` and reject `NaN`.
[...]

Issue #266, opened 2026-08-05
Document confidence scores and make OCR routing thresholds configurable


yulinlina:
## Description

`pdf-inspector` classifies PDFs and returns a confidence score (`0.0-1.0`) plus per-page OCR routing recommendations. However, the README and binding docs do not explain how to interpret these confidence values or what thresholds are used to decide that a page/document should be treated as `TextBased`, `Scanned`, `ImageBased`, or `Mixed`.

## Motivation

Users adopting this library for OCR routing need predictable behavior:

- A confidence of `0.65` may be acceptable for one workflow but too low for another.
- Applications may want to favor avoiding OCR costs, while others may want to avoid false `TextBased` classification.
- Without documented defaults, users cannot easily tune precision/recall or understand surprising results.

Clarifying and exposing threshold behavior would make the library easier to adopt in Python, Node.js, and WASM integrations.

## Proposed Solution

1. Add a documentation section, ideally in the README and language-specific docs, explaining:
   - What the document-level confidence score represents.
   - How per-page routing decisions are derived.
   - Default thresholds and behavior for `Mixed` PDFs.
   - Recommended threshold presets for common use cases, such as "avoid OCR" vs "maximize text extraction reliability".

2. Optionally expose configurable thresholds in the public API, for example:
   - `min_text_confidence`
   - `min_scanned_confidence`
   - `page_routing_strategy` or similar

3. Provide short examples for Rust, Python, Node.js, and WASM showing how to override thresholds and inspect per-page routing.

## Alternatives Considered

- Document the current behavior only, without API changes.
- Leave thresholds hardcoded and require users to post-process confidence scores themselves.
- Provide separate helper utilities for thresholding instead of core configuration.

---
🤖 *Suggested by [GitHub Trend-Hacking Pipeline](https://github.com/yulinlina/start_rise)*

AbarnaaSree:
Hi! I'd like to work on this issue. Is it okay if I take this up?

yulinlina:
Yes, go ahead — this is a good issue to take.

To make the PR easy to review, I’d suggest splitting it into two pieces:

1. **Documentation**
   Add a short section in the README and binding docs explaining:
   - what the document-level confidence score means,
   - how page-level routing decisions are made,
   - what the current default thresholds are,
   - and how `Mixed` is chosen when pages disagree.

2. **Configurable thresholds**
   Introduce an optional config type so callers can override the defaults without changing existing behavior. Something like:

```rust
[...]

PR #454, opened 2026-08-22
feat(go): add Go bindings via cgo

hemenduroy:
## Summary

Adds a Go binding (`go/`), matching the document-processing surface of the `napi` (Node) and Python bindings — including selective OCR — via a hand-written C ABI + [cgo](https://pkg.go.dev/cmd/cgo).

- **API parity**: `Classify`, `ExtractText`, `ProcessPdf`, `DetectPdf`, `ExtractPagesMarkdown`, `ExtractTextWithPositions`, `ExtractStructureElements`, `ExtractTextInRegions`, `ExtractTablesInRegions`, `DetectVectorGridInRegion`, `ExtractTablesWithStructure` (+`Cells`/+`Auto`), and `ProcessPdfWithOcr`. This matches Node's full surface and exceeds Python's (Python doesn't expose the region/table-structure-recovery functions).
- **ABI design**: two argument shapes cover all 13 operations — `(data, len)` for the two that take no options, `(data, len, params_json)` for everything else — rather than a bespoke C parameter list per function. Every call returns a JSON envelope (`{"ok":true,"result":{...}}` / `{"ok":false,"error":"..."}`), the same pattern `napi/src/lib.rs` uses (result structs + `catch_unwind` on every entry point), adapted for a language with no napi/PyO3-equivalent marshaling.
- **OCR**: `go/Cargo.toml` builds against the core crate's `ocr` feature, same as `napi/Cargo.toml`/`pyproject.toml`. PDFium/ONNX Runtime stay dynamically loaded at runtime (not linked at build time) — same as every other binding.
- **Distribution**: `go generate` downloads a prebuilt native library from a `go/vX.Y.Z` GitHub Release for darwin/linux × arm64/x64 (checksum-verified), falling back to `cargo build --release` (`make native`) elsewhere. `.github/workflows/publish-go.yml` cross-builds those release artifacts whenever `go/Cargo.toml`'s version changes.
- **CI**: new `go` job in `ci.yml` — builds/lints/tests from source on Linux+macOS on every push/PR, `go test -race`.
- **Tests**: 63 tests covering every function against real fixtures, password-protected PDFs (via `ProcessPdfWithOcr(..., Mode: OcrOff)`), concurrent-caller safety under the race detector, and a corpus smoke test against every fixture in `tests/fixtures/`.

## Relationship to #292

This overlaps in intent with #292 (`feat(go): add pure Go bindings (CGO=0) via WebAssembly and wazero`), which takes a different architecture — compiling to `wasm32-wasip1` and running it with [wazero](https://github.com/tetratelabs/wazero), avoiding cgo entirely. I want to flag that explicitly rather than have this look like an accidental duplicate. The trade-offs as I see them:

- **wazero's real advantage**: `CGO_ENABLED=0` — no C toolchain, no per-platform native binaries, single embeddable `.wasm` blob. Genuinely valuable for Go teams that avoid cgo on principle.
- **Why I still think cgo is the better fit here**: the core crate disables `rayon` (parallel parsing) for every `wasm32` target, so a WASM-based binding inherits the same single-threaded ceiling as the existing browser binding — a real throughput cost on large documents that cgo doesn't pay. More significantly, OCR (PDFium + ONNX Runtime, loaded via native `dlopen`) is architecturally impossible from inside a WASM sandbox — a wazero-based binding can never grow into OCR support, which is a real limitation given `processPdfWithOcr` just shipped for Node/Python (#405) and OCR routing is clearly an active investment area (#266, #267). cgo also matches the distribution pattern already proven out for `napi`/Python (per-platform native binaries + release matrix) rather than introducing a third paradigm.

Happy to have maintainers weigh in on which direction is preferred for the project's Go story — these aren't mutually exclusive if there's appetite for both (e.g. wazero for a "no toolchain, no OCR" tier and cgo for full parity), but I wanted the comparison on the table rather than silently competing.

## Test plan

```bash
cd go && cargo fmt --check && cargo clippy --release -- -D warnings && cargo build --release
cd pdfinspector && go generate ./... # or: cargo build --release in go/, then skip this
go vet ./... && go test -race ./... -v
```

All 63 tests pass; verified against a from-scratch `cargo clean` rebuild. Root-level `cargo fmt --check`, `cargo clippy -- -D warnings`, and `cargo test` all pass unaffected (no core crate changes).

---
## Summary by cubic
Adds cgo-based Go bindings that mirror the Node/Python document-processing API, including selective OCR, giving Go users full parity with the Rust core. Behavior changes: page filters distinguish nil vs empty (an explicit empty slice processes zero pages); `ProcessPdf` and `ExtractTextWithPositions` use 1-indexed pages; `ExtractPagesMarkdown` returns 1-indexed summary fields while its pages argument remains 0-indexed.

- C ABI returns JSON envelopes; two signatures cover the surface: (data, len) for no-option calls and (data, len, params_json) for everything else. Panics are caught at the FFI boundary; vector‑grid detection requires all parameters.
- Builds the `go` crate with the `ocr` feature; PDFium and ONNX Runtime load dynamically at runtime. CI builds the Go native library and runs an OCR pass to verify dynamic loading.
- `go/pdfinspector` uses `go generate` to fetch prebuilt native libraries via `pdfinspector/internal/fetchnative`; musl Linux rejects glibc‑only prebuilts and falls back to local builds. Downloads are checksum‑verified, never overwrite existing files, and write a `.fetchnative-release.json` marker to track the installed release and avoid staleness. README clarifies artifact sets are platform‑specific.
- `.github/workflows/publish-go.yml` cross‑builds and uploads per‑platform artifacts to `go/pdfinspector/vX.Y.Z` releases with scoped permissions; re‑runs are idempotent.
- Tests cover the full surface, password‑protected PDFs, concurrent callers under `go test -race`, a corpus smoke test, and regressions for page indexing, explicit zero‑page selection, and fetchnative staleness; docs now call out `ExtractPagesMarkdown`’s 1‑indexed result fields.

**Migration**
- Use 1‑indexed pages for `ProcessPdf` and `ExtractTextWithPositions`.
- Treat `ExtractPagesMarkdown` summary fields as 1‑indexed.
- Pass an explicit empty slice to process zero pages.
- Run `go generate ./...` (especially after upgrading `go/pdfinspector`) or build once locally; ensure OCR dependencies are available at runtime when using OCR modes.

<sup>Written for commit fcf3a3709aa722f70a9fb9d94a0874a3cd2384e1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/pdf-inspector/pull/454?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


hemenduroy:
@cubic-dev-ai

hemenduroy:
@cubic-dev-ai

hemenduroy:
@cubic-dev-ai

hemenduroy:
@cubic-dev-ai

hemenduroy:
@cubic-dev-ai

hemenduroy:
@cubic-dev-ai review

hemenduroy:
@cubic-dev-ai review

hemenduroy:
Ready for review

thiagogcm:
Hi!

I've been experimenting with Java Bindings via FFM API. Can we have a dedicate C layer to enable other languages bindings through a common ABI?
still work in progress but works (https://github.com/thiagogcm/pdf-inspector)

hemenduroy:
> Hi!
> 
> I've been experimenting with Java Bindings via FFM API. Can we have a dedicate C layer to enable other languages bindings through a common ABI? still work in progress but works (https://github.com/thiagogcm/pdf-inspector)

Not too sure whether this repo even accepts contributions but I put up this PR nonetheless 🤔 

I looked but couldn’t find any contribution guidelines.

PR #476, opened 2026-08-31
fix(markdown): recognize extractor bullet glyphs (▪ ‣ ◆) as list items

### LINKED ISSUE #475: Markdown list detection only recognizes 6 bullet glyph types; extractor modules already handle 10+, so ▪/‣/◆ list items are silently demoted to plain paragraphs
## Summary

`is_list_item()` and `format_list_item()` in `src/markdown/classify.rs` only recognize a small set of bullet glyphs (`• ○ ● ◦` plus `- *`), while other modules in the same crate recognize a much wider set. A list item like `▪ Item` is therefore not detected as a list at all and is emitted as a plain paragraph.

## Bullet glyph sets across modules (as of current `main`)

| Module | Glyphs | Purpose |
|---|---|---|
| `src/markdown/classify.rs:73-93` (`is_list_item`, `starts_with_bullet_marker`) | `• ○ ● ◦ - *` | Decides whether a line is a list item in markdown output |
| `src/markdown/classify.rs:130` (`format_list_item`) | `• ○ ● ◦` | Converts the marker to a markdown `-` |
| `src/extractor/layout.rs:798` (`is_list_marker_column`) | `• ● ○ ◦ ▪ ▫ ◆ ◇ ■ □` | Rejects spurious column-break candidates on bullet columns |
| `src/extractor/underline.rs:400` (`is_bare_list_marker`) | `• ◦ ▪ ▫ ‣ ⁃ ● ○ ■ □ - *` | Excludes bare markers from underline/strikethrough detection |
| `src/extractor/mod.rs:637` (`is_standalone_bullet_text`) | `• ○ ● ◦` | Merges standalone bullet glyphs with following text |
| `src/tables/mod.rs:1319` | `• ● ·` | Numeric-cell heuristics for table detection |
| `src/vision/fusion.rs:763` | `• ● ○ ◦ ▪ - —` | Vision fusion bullet set |
| README "Markdown output" table | `• - * ○ ● ◦` | Documented behavior |

The extractor knows that `▪ ▫ ◆ ◇ ■ □ ‣ ⁃ ·` are list markers (it refuses to split bullet columns on them, and excludes them from underline detection), but the markdown layer never receives that knowledge.

## Reproduction

Verified on current `main` (added a scratch unit test, since removed):

```rust
is_list_item("▪ Item")                // false  ← expected: true
starts_with_bullet_marker("▪ Item")    // false  ← expected: true
format_list_item("▪ Item")             // "▪ Item"  ← expected: "- Item"
is_list_item("‣ Item")                 // false  ← expected: true
is_list_item("◆ Item")                 // false  ← expected: true
```

So a PDF list rendered with `▪` markers produces:

```markdown
▪ First item
▪ Second item
▪ Third item
```

instead of:

```markdown
- First item
- Second item
- Third item
```

Downstream consequence: since `is_list_item()` gates heading classification (`src/markdown/heading.rs:291`) and code classification, a short bold `▪`-bullet line may additionally be misclassified as a heading.

## Proposed direction

Add a single shared bullet-marker constant (e.g. `pub(crate) const BULLET_MARKERS: &[char]` in `src/markdown/classify.rs` or a common module) and use it in `is_list_item()`, `starts_with_bullet_marker()`, `format_list_item()`, `is_list_marker_column()`, `is_bare_list_marker()`, and `is_standalone_bullet_text()`, so glyph sets can no longer drift apart between modules. The extractor sets suggest the full working set is at least:

```text
• ● ○ ◦ ▪ ▫ ◆ ◇ ■ □ ‣ ⁃ · - *
```

`·` (U+00B7 MIDDLE DOT) is intentionally ambiguous (it also appears in normal prose), so it may warrant separate treatment rather than inclusion in the automatic list-marker set.

## Environment

- pdf-inspector `main` (checked out today)
- Only library-level unit tests; no specific PDF needed to reproduce (works from pure string inputs)

shahidbeig-a11y:
Fixes #475

---
## Summary by cubic
Recognizes additional bullet glyphs (▪ ‣ ◆ and related) as list items in markdown extraction, so PDF lists using them are no longer emitted as plain paragraphs or misclassified as headings. Fixes #475.

- Shares `BULLET_GLYPHS` across list detection, formatting, standalone bullet merging, and the layout marker guard.
- Keeps `-` and `*` as separate checks and excludes middle dot, en dash, and em dash.
- Expands tests covering extended glyphs, negative cases, heading rejection, and a square-bullet merge stream-order regression.

<sup>Written for commit 87d8478863c37b9a3c83ca2143b5438d5e6dc916. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/pdf-inspector/pull/476?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/extractor/layout.rs
+++ b/src/extractor/layout.rs
@@ -790,25 +790,18 @@ fn find_relative_valleys(
 }
 
 /// Detect whether a side of a gutter consists predominantly of list-marker
-/// glyphs (•, ●, ○, ◦, ▪, ▫, ◆, ◇). A column of bullets on the left margin
-/// creates a spurious histogram valley between the bullet and the content.
-/// Treating it as a real column splits each list item's text across two
-/// "columns," so we reject these candidates.
+/// glyphs from [`crate::markdown::classify::BULLET_GLYPHS`]. A column of bullets
+/// on the left margin creates a spurious histogram valley between the bullet
+/// and the content. Treating it as a real column splits each list item's text
+/// across two "columns," so we reject these candidates.
 fn is_list_marker_column(items: &[&&TextItem]) -> bool {
-    const LIST_MARKERS: &[char] = &['•', '●', '○', '◦', '▪', '▫', '◆', '◇', '■', '□'];
+    use crate::markdown::classify::is_standalone_bullet_glyph;
     if items.is_empty() {
         return false;
     }
     let marker_count = items
         .iter()
-        .filter(|i| {
-            let t = i.text.trim();
-            let mut chars = t.chars();
-            match (chars.next(), chars.next()) {
-                (Some(c), None) => LIST_MARKERS.contains(&c),
-                _ => false,
-            }
-        })
+        .filter(|i| is_standalone_bullet_glyph(&i.text))
         .count();
     // Require ≥80% of items on this side to be standalone markers. A handful
     // of non-marker items (stray page numbers, footnote refs) shouldn't
--- a/src/extractor/mod.rs
+++ b/src/extractor/mod.rs
@@ -634,12 +634,7 @@ fn effective_merge_width(item: &TextItem) -> f32 {
 }
 
 fn is_standalone_bullet_text(text: &str) -> bool {
-    let trimmed = text.trim();
-    let mut chars = trimmed.chars();
-    matches!(
-        (chars.next(), chars.next()),
-        (Some(c), None) if crate::markdown::classify::BULLET_GLYPHS.contains(&c)
-    )
+    crate::markdown::classify::is_standalone_bullet_glyph(text)
 }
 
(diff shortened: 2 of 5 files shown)


shahidbeig-a11y:
Follow-up: chart-padding guard test now iterates `BULLET_GLYPHS` end-to-end (0fb69bb) so the pin stays in sync with the shared glyph set.

PR #479, opened 2026-09-02
fix: recognize invisible-layer OCR text under scanned images with sparse visible chrome

Plaba:
## Problem

PDFs from "searchable image" scanning tools (common in government FOIA releases, e.g. Adobe's OCR-and-embed workflow) place the real page text as an invisible layer (`Tr 3`) directly under the full-page scan raster, often alongside a small amount of genuinely visible text — a bates stamp, document ID, or page number. `pdf-inspector`'s lightweight classifier didn't account for text-rendering mode at all, so these pages were misclassified two different ways depending on how much invisible text was present:

- With a substantial invisible layer, its `Tj`/`TJ` operator count alone cleared the text-ops floor, so the page classified `TextBased` — `extract_pages_markdown`/`process_pdf`/`classify_pdf` all silently returned only the small visible stamp and dropped the actual body content.
- Once invisible text is correctly excluded from that count (this PR), a page with no *visible* text left over classifies `ImageBased` instead — which routes to a fresh real-OCR render rather than recovering the existing, higher-fidelity embedded layer.

Neither is `Mixed`, which is the only classification whose recovery path retries extraction with invisible text included.

## Root cause

`analyze_page_content`'s lightweight byte-scanner (`scan_content_for_text_operators`) counted every `Tj`/`TJ` toward `text_operator_count` with no awareness of the current text-rendering mode. Two further issues compounded this once mode-tracking was added:

1. A hardcoded `unique_alphanum_chars < 10` gate in `pages_with_template_images` rejected pages whose visible chrome happened to be diverse (a bates stamp like `EFTA01234567` has more than 10 unique alphanumeric characters) will block `Mixed` classification even with invisible-text tracking in place.
2. The `Tr` operator itself can appear immediately adjacent to the next operator with no whitespace (`3 Tr/F0 12 Tf...`), the same concatenation pattern already tolerated for `Tf`. The initial fix's follower check only accepted whitespace, silently missing this in real-world files.

## Fix

- `scan_content_for_text_operators` now tracks text-rendering mode (`Tr`) as a small piece of state, correctly scoped by `q`/`Q` graphics-state save/restore. `Tj`/`TJ` shown under mode 3 (invisible) no longer count toward `text_operator_count` or contribute to `unique_alphanum_chars`, and are instead reported via a new `has_invisible_text` signal on `PageAnalysis`.
- `pages_with_template_images` and the document-level classification branch now also route to `Mixed` when a page has confirmed invisible text, even if its (now-excluded) visible-only stats would otherwise fail the alphanum/text-ops floor.
- `is_word_end` now treats PDF delimiter characters (not just whitespace) as valid token boundaries, fixing both the `3 Tr/F0` adjacency case and a `q%comment` case that could desync the new `q`/`Q` render-mode stack.

No changes to the invisible-text *recovery* path itself (`extract_text_in_regions_mem`'s `include_invisible` retry, the OCR pipeline's native-candidate fusion) — this PR only fixes the classification signal that gates whether those existing mechanisms run at all.

## Testing

- 9 new unit tests in `src/detector.rs`: byte-level coverage of the `Tr` state machine (basic exclusion, mixed visible/invisible runs, reset, `q`/`Q` scoping incl. 2-deep nesting, both delimiter-adjacency bugs) plus 2 classification-level tests using a new `masthead_scan_page_with_invisible_body` helper.
- `tests/fixtures/scan_with_invisible_ocr_layer.pdf`: a synthetic fixture (not a real document — built with placeholder names/content) reproducing the exact real-world pattern: a rasterized scan of a business letter, a small visible document-ID stamp landing exactly on the previous alphanum boundary, and the same letter text as an invisible OCR layer underneath.
- Integration test in `tests/integration_tests.rs` verifying `classify_pdf`/`extract_pages_markdown` agree the page needs OCR/recovery (no OCR runtime required).
- Integration test in `tests/ocr_tests.rs` verifying the full `process_pdf_with_ocr` pipeline recovers the invisible layer and fusion prefers it over OCR's own (much weaker) re-derivation of the raster.
- Full suite: 1127 lib tests, 166 integration tests, 6 OCR pipeline tests, all passing. `cargo fmt --check` clean.

---
## Summary by cubic
Fixes classification of searchable-image PDFs that contain a full-page raster, a small visible stamp, and an invisible OCR layer. Previously, the invisible text could make a page `TextBased` and drop the body, or `ImageBased` and trigger fresh OCR; now visible text drives classification while detected invisible text routes the page through `Mixed` recovery.

- Tracks `Tr` render mode with `q`/`Q` scoping across page streams and form XObjects.
- Excludes invisible `Tj`, `TJ`, and `'` text from visible counts while reporting non-empty invisible text separately.
- Queues matching pages for recovery even when visible chrome is diverse.
- Handles PDF delimiter boundaries and ignores operators inside strings and comments.

**Testing**

- Adds scanner (incl. `'` operator and its delimiter-adjacency), classification, multi-stream, routing, and fixture-based integration coverage.
- Verifies OCR fusion prefers recovered native text over a fresh OCR pass.

<sup>Written for commit 09f145ff5521753ae81f6a6050574ad0d7d076c8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/pdf-inspector/pull/479?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Plaba:
I went through all 7 findings directly (no sub-agents needed — I already had full context on this code from writing it, and the issues were tightly scoped to specific functions). Six were valid and are now fixed; one is incorrect and I left it as-is.

**Fixed (6 valid findings):**

1. **`ocr_tests.rs` P2** — the OCR-fusion test never asserted OCR actually ran, so it could pass without exercising the behavior it claimed to test. Added `assert_eq!(result.pages_routed_to_ocr, vec![1])`.
2. **`detector.rs` P1 (Phase-2 routing gap)** — the Phase-1 aggregate classifier correctly routed a page to `Mixed` when it had an invisible layer, but the *per-page* OCR-queueing logic (`detect_from_document`'s Mixed loop, and its duplicate in `page_ocr_signals`) never checked `has_invisible_text` at all. A page with 10–49 diverse visible ops plus an invisible body would classify the whole doc `Mixed` but never get queued for recovery. Fixed both call sites; added a regression test that specifically targets that 10–49 range (my original tests only had 1 visible op, which accidentally cleared a different floor and masked this gap).
3. **`detector.rs` P1 (cross-stream state)** — `render_mode`/the `q`/`Q` stack were function-local, reset on every call, but a page's `/Contents` array is logically one continuous stream per PDF 32000-1 §7.8.2 — the split may only fall between complete tokens, so state must persist across it. Threaded both as caller-owned state through `analyze_page_content`'s loop over the array. (Form XObjects get their own fresh local state instead, correctly — they're implicitly bracketed by their own `q`/`Q` per §8.10.)
4. **`detector.rs` P2 (empty operand)** — `has_invisible_text` fired on an empty `() Tj` or numeric-only `TJ` array with nothing recoverable. Now gated on the operand actually containing non-whitespace string content.
5. **`detector.rs` P1 (apostrophe operator)** — `'` (show-text-and-advance-line) was never recognized at all, visible or invisible, despite the function's own doc comment claiming it was handled. Added it with the same render-mode/emptiness/preceding-closer checks as `Tj`.
6. **`detector.rs` P2 (Tr inside strings/comments)** — `Tr`'s bare-number operand has no natural guard against matching inside `(a string containing 3 Tr)` or `% a comment with 3 Tr`, unlike `Tj`/`TJ` which lean on a preceding string/array closer. Added proper forward string/comment-span tracking to the scanner, closing this for every operator it looks for, not just `Tr`.

**Not fixed — I believe it's incorrect:**

Violation #6, "`q`/`Q` must not restore `Tr`": this contradicts the PDF spec. ISO 32000-1 Table 52 ("Device-Independent Graphics State Parameters") explicitly lists **text state** (which includes text rendering mode) as one of the parameters saved/restored by `q`/`Q` — it's part of the graphics state, not exempt from it. This is also why `q 3 Tr ... Q` is such a common real-world idiom for scoping a temporary invisible/stroke run without an explicit `0 Tr` reset afterward — it only works because `Q` restores the prior mode. I kept the stack-based save/restore as originally implemented and tested (`test_tr_scoped_by_q_and_restored_on_q_pop`, `test_tr_nested_q_stack_restores_each_level`). Happy to dig further if you have a specific counter-example in mind, but I'd want to see it before reversing this.

Full suite after all fixes: 1136 lib tests, 166 integration, 6 OCR pipeline tests — all passing, `cargo fmt --check` clean.

PR #489, opened 2026-09-02
test(bindings): route encrypted fixture through password-aware Python API

abimaelmartell:
## Summary

`TestMultipleFixtures.test_process_all_fixtures` in `tests/test_python.py` runs every PDF in `tests/fixtures/` through `process_pdf`, which has no password parameter, so the encrypted fixture added with `--password` support failed with `ValueError: PDF is encrypted`. CI does not run pytest, so the failure went unnoticed.

- Encrypted fixtures are now routed through `process_pdf_with_ocr` with `mode="off"` and their known password, looked up from a small `ENCRYPTED_FIXTURE_PASSWORDS` map. OCR stays off, so no external runtime is needed.
- Adds two focused tests: `process_pdf` on the encrypted fixture raises `ValueError` mentioning encryption, and the path and bytes OCR entry points decrypt it with the password and return the fixture's real text, mirroring the existing Rust integration test.

## Verification

Built the wheel with `maturin build --release` into a Python 3.12 venv and ran `pytest tests/test_python.py`: 86 passed. The unmodified test file fails on the encrypted fixture against the same wheel.

---
## Summary by cubic
Updates Python fixture tests so encrypted PDFs use password-aware `process_pdf_with_ocr` instead of `process_pdf`, which cannot accept passwords. OCR stays off, avoiding external runtime dependencies while allowing the full fixture sweep to pass.

- Adds a filename-to-password map as the single source of truth for encrypted fixtures.
- Covers the missing-password error and password-based path and bytes APIs, checking decrypted fixture text.
- `pytest tests/test_python.py` passes with 86 tests.

<sup>Written for commit befa71fe4f85f8f2f5d40ede7551d25d92033405. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/pdf-inspector/pull/489?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/tests/test_python.py
+++ b/tests/test_python.py
@@ -16,6 +16,11 @@ def fixture_bytes(name: str) -> bytes:
         return f.read()
 
 
+# Fixtures that need a user password to open. `process_pdf` has no password
+# parameter, so these are exercised through `process_pdf_with_ocr`, which does.
+ENCRYPTED_FIXTURE_PASSWORDS = {"encrypted-secret123.pdf": "secret123"}
+
+
 # ---------------------------------------------------------------------------
 # process_pdf
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


abimaelmartell:
Review fix is pushed (the focused test now reads the password from `ENCRYPTED_FIXTURE_PASSWORDS`). Please re-review this PR. @cubic-dev-ai

PR #496, opened 2026-09-04
feat(vision): selectable pinned OCR model sets with a PP-OCRv5 Korean set

NUMORDO:
## Summary

Adds `OcrModelSet` so the selective-OCR path can load a different pinned, checksum-verified model set without changing anything else in the pipeline. The default set (`PpOcrV6Small`) and all existing behavior are unchanged; the new `PpOcrV5Korean` set makes Hangul readable.

Motivation: the PP-OCRv6 Small dictionary contains no Hangul at all (18,709 entries, 0 in U+AC00–U+D7A3), so every Korean scanned page currently comes back empty even though the routing, rendering, fusion and Markdown assembly all work. The PP-OCRv5 Korean recogniser published with `oar-ocr` (already the crate's OCR backend) covers all 11,172 Hangul syllables plus Latin letters and digits, and pairs with the script-agnostic PP-OCRv5 mobile detector. Together the set is ~18 MB, smaller than the default.

## What changed

- `vision::OcrModelSet` (`PpOcrV6Small` default, `PpOcrV5Korean`) with stable identifiers, a lenient `parse`, and `Display`. `#[non_exhaustive]` so more sets can follow.
- `OcrOptions::model_set(...)` / `OcrOptions.model_set`. `Default` still selects PP-OCRv6 Small.
- `vision::PP_OCR_V5_KOREAN` manifest pinned to the `oar-ocr` v0.3.0 release files (SHA-256 and exact sizes), and `OcrModelSet::manifest()`.
- `cached_ocr_engine` now resolves, downloads, and caches by the selected manifest. The engine cache key already carried manifest id/revision/digests, so switching sets replaces the loaded sessions instead of mixing artifacts; each set also lives under its own `<id>/<revision>` cache directory.
- CLI: `pdf2md --ocr-model-set pp-ocrv5-korean` (listed with the other OCR-only options; invalid values print the accepted identifiers).
- Docs: `docs/ocr-runtime.md` "Model sets" table and examples, `docs/rust-api.md`, CHANGELOG.

Python and Node bindings are untouched in this PR (they keep the default set); happy to add a `model_set` option to both in a follow-up if you want parity in the same release.

## Verification

- `cargo fmt`, `cargo clippy -- -D warnings`, `cargo clippy --features ocr -- -D warnings`, `cargo test`, `cargo test --features ocr --lib vision::` on the pinned 1.98.0 toolchain.
- New unit tests: Korean manifest validates and has the three expected artifact kinds; every `OcrModelSet` resolves to a manifest whose `id` matches; sets use distinct cache roots; identifier parsing round-trips and rejects unknown values; `OcrOptions` default is unchanged.
- End-to-end (Linux x64, PDFium `native-v7988`, ONNX Runtime 1.27.0, CPU): a Korean government PDF rasterised to a 150 dpi image-only PDF (3 pages, classified `scanned`) processed with `--ocr auto --ocr-model-set pp-ocrv5-korean --ocr-offline --ocr-model-dir …`. Results are in the comment below.

## Notes for reviewers

- Artifact digests were computed from the files downloaded from the `oar-ocr` v0.3.0 GitHub release (`pp-ocrv5_mobile_det.onnx` 4,826,518 B, `korean_pp-ocrv5_mobile_rec.onnx` 13,446,374 B, `ppocrv5_korean_dict.txt` 47,451 B).
- No new dependencies. No change to the default feature set or to the text-only path.

---
## Summary by cubic
Adds `OcrModelSet` so the selective-OCR path can load a different pinned, checksum-verified model set. The default `PpOcrV6Small` set is unchanged, and the new `PpOcrV5Korean` set makes Hangul readable — Korean scanned pages previously came back empty because the default dictionary has no Hangul.

- `PpOcrV5Korean` pairs the script-agnostic PP-OCRv5 mobile detector with the Korean recogniser and dictionary (~18 MB total).
- `OcrOptions::model_set(...)` and `pdf2md --ocr-model-set pp-ocrv5-korean` select the set; the engine cache is keyed by the selected set, so switching sets never mixes artifacts.
- `OcrOptions` gained a public `model_set` field, so struct literals that spell out every field need the new field.
- Python and Node bindings are untouched and keep the default set.

<sup>Written for commit 76cdc11230394cdbd04eb1a8c1774e7310730775. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/pdf-inspector/pull/496?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,16 @@ version and date. Earlier releases are described in their
 
 ### Added
 
+- `OcrModelSet`: selects which pinned, checksum-verified model set the OCR
+  engine loads. `PpOcrV6Small` is the default and unchanged; the new
+  `PpOcrV5Korean` set pairs the script-agnostic PP-OCRv5 mobile detector with
+  the Korean PP-OCRv5 mobile recogniser and its dictionary (all 11,172 Hangul
+  syllables plus Latin letters and digits, ~18 MB in total). Exposed as
+  `OcrOptions::model_set(...)`, the `PP_OCR_V5_KOREAN` manifest, and
+  `pdf2md --ocr-model-set pp-ocrv5-korean`. Each set resolves, downloads, and
+  caches under its own manifest id and revision, and the in-process engine
+  cache is keyed by the selected set, so switching sets never mixes artifacts.
+
 - `TextItem::baseline_shift`: signed offset, in points, of a superscript or
   subscript glyph run from the baseline of the body text it is attached to
   (positive = raised, negative = lowered, `0` for normal text). Exposed as
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -113,6 +113,11 @@ version and date. Earlier releases are described in their
 
 ### Changed
 
+- `OcrOptions` gained the public `model_set` field. Construct it through
+  `OcrOptions::new()`/`Default` and the builder methods, as the docs show;
+  a struct literal that spells out every field needs the new field. This
+  follows the crate's existing practice for public structs (`TextItem`
+  gained `rotation`, `advance_known`, and `baseline_shift` in this cycle).
 - **Coordinate frame of positioned output — consumer action may be required.**
   `extract_text_with_positions*` (Rust), `extractTextWithPositions` (Node),
   `extract_text_with_positions[_bytes]` (Python) and `pdf2md --items-json` now
--- a/docs/ocr-runtime.md
+++ b/docs/ocr-runtime.md
@@ -17,6 +17,37 @@ The reproducible runtime path uses these builds:
 Use these versions for the reproducible path. Other compatible shared-library
 builds may work, but are not part of the release smoke test.
 
+## Model sets
+
+The recogniser and its character dictionary decide which scripts OCR can
+read; detection is script-agnostic. `OcrModelSet` selects one pinned,
+checksum-verified set. The default is unchanged.
+
+| `OcrModelSet` | Identifier | Files (revision) | Scripts | Download size |
+|---|---|---|---|---|
+| `PpOcrV6Small` (default) | `pp-ocrv6-small` | `pp-ocrv6_small_det.onnx`, `pp-ocrv6_small_rec.onnx`, `ppocrv6_dict.txt` (`oar-ocr-v0.7.0`) | Latin scripts, Simplified and Traditional Chinese, Japanese | about 31 MB |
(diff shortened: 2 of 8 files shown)


NUMORDO:
End-to-end measurement for the Korean set (Linux x64, 12-thread laptop CPU, release build with `--features ocr`, PDFium `native-v7988`, ONNX Runtime 1.27.0, offline model directory):

Input: a Korean government legal-interpretation PDF (text-based) rasterised to a 150 dpi greyscale image-only PDF, 3 pages, classified `scanned` with `pages_needing_ocr = [1, 2, 3]`. Ground truth = the original text layer of the same pages, extracted with `pdf2md`.

| Run | Result |
|---|---|
| `--ocr auto --ocr-model-set pp-ocrv5-korean --ocr-offline --ocr-model-dir …` | all 3 pages routed to OCR, 1,434 Hangul characters recognised; character-level recall **0.918**, precision **0.916** against the text layer (whitespace/markup-normalised `difflib` matching, 1,647 reference characters) |
| same input, default `pp-ocrv6-small` | 0 Hangul characters (176 characters of non-Korean output), as expected from the dictionary |
| wall time (3 pages, warm cache) | 32.6 s wall / 93.7 s user with the Korean set; 16.7 s wall with the default set on the same pages. `--ocr-dpi 100/120` did not change either time or accuracy materially, so the cost is recognition of many short lines (dense legal text), not rendering |

Clean text PDFs are unaffected (no renderer, model, or ORT initialisation happens before routing).

NUMORDO:
Re the automated note on `OcrOptions.model_set`:

- `OcrOptions` is documented and constructed only through `OcrOptions::new()` / `Default` and the builder methods (`docs/rust-api.md`, `docs/ocr-runtime.md`, the CLI and the Python bindings all use that path); there is no struct-literal construction of it anywhere in the repo.
- Adding a public field to a public struct in a minor release is the crate's existing practice: in this same unreleased cycle `TextItem` gained `rotation` and `advance_known` (#486) and `baseline_shift` (#488).
- Any way of attaching a new option to `OcrOptions` (public or private field) has the same effect on a fully spelled-out literal, so the choice is between following the existing practice or introducing a parallel options type. I kept the field and added a note under `Changed` in the CHANGELOG (pushed) so the release notes call it out.

If you would rather gate this behind `#[non_exhaustive]` on `OcrOptions` (which would also stop literals, but makes the policy explicit for the future) or a different shape, tell me and I will adjust.

PR #549, opened 2026-09-19
test(positions): cover the boxes of text drawn with a reflected matrix

### LINKED ISSUE #515: Positioned boxes shift left by the full text width for upright negative-font-size text
## Problem

In `pdf-inspector==1.18.0`, an upright `/Rotate 0` page whose text uses a negative `Tf` size returns positioned boxes off the left edge of the page.

The visible text starts at `x=72` on a 612-point-wide page. Both public positioned-text APIs instead return a same-width box ending at `x=72`: the first item is `x=-156.768`, `width=228.768`; the second is `x=-82.080`, `width=154.080`.

This makes a viewer highlight or region lookup miss text that is visibly on the page.

## Environment

- `pdf-inspector==1.18.0`
- CPython 3.13 on macOS arm64
- Public fixture: https://github.com/run-llama/liteparse/blob/python-v2.14.4/integration_tests_data/negative_font_size.pdf
- Fixture SHA-256: `2bdf661e6dec41b1802e79e8added8ca939821885881b9038810e1a5748b307d`

## Reproduction

Install `pdf-inspector==1.18.0`, save the linked public fixture as `negative_font_size.pdf`, then run:

```python
from pathlib import Path
import pdf_inspector as p

data = Path("negative_font_size.pdf").read_bytes()
items = p.extract_text_with_positions_bytes(data)
with_rotations = p.extract_text_with_positions_and_rotations_bytes(data)

for item in items:
    print(item.text, item.x, item.y, item.width, item.height, item.rotation, item.font_size)
print("page_rotations:", with_rotations.page_rotations)
print("same items:", [(x.text, x.x, x.width) for x in items] == [
    (x.text, x.x, x.width) for x in with_rotations.items
])
```

Actual output on 1.18.0:

```text
Upright text drawn with a negative font size -156.76800537109375 720.0 228.76800537109375 12.0 0.0 12.0
Second line stays upright too -82.08000183105469 692.0 154.0800018310547 12.0 0.0 12.0
page_rotations: []
same items: True
```

## Expected

The fixture visibly renders both lines upright beginning at `x=72`. Poppler's independent `pdftotext -bbox` output reports `xMin=72.000000` for each line, with the first ending at `xMax=300.768000` and the second at `xMax=226.080000`.

The positioned API should return the axis-aligned visible glyph bounds, so the expected starts are `x=72` with the same widths. The page has `MediaBox [0 0 612 792]` and `/Rotate 0`; no CropBox or page-turn transform is involved.

## Scope

This is a focused geometry defect, not a claim that the 1.18.0 coordinate-frame changes are broadly incorrect. The public CropBox fixture introduced in 1.18.0 behaves as documented in a separate local check.

PR #486 added rotated-run geometry and handling for negative `Tf` text. It is a relevant change to inspect, but this report does not assume the exact root cause or prescribe a fix.

The fixture is public and fictional. It contains only these two visible sentences and no personal metadata, attachments, JavaScript, forms, encryption, customer content, or internal paths.

abimaelmartell:
Regression coverage for #515. Text drawn with a negative `Tf` size, a negative `Tz` scale or a turning text, page or form matrix reports its box where its glyphs are: reflections that cancel into upright text give the box of the same string drawn upright, from the origin rather than one text width to its left, and reflections that stand give the box their glyphs occupy. The shape reported in the issue, a negative size and scale under a y-flipping page matrix, was corrected by the horizontal scaling fix in 1.19.0; this pins every shape down on the positioned-text APIs and the region API with synthetic pages, so it cannot come back.

Closes #515.

---
## Summary by cubic
Adds synthetic-PDF regression coverage for #515, ensuring positioned-text and region extraction report glyph boxes correctly for negative `Tf`/`Tz` values and reflected text, page, and form matrices. It covers both reflections that cancel to upright text and reflections that remain visible.

- Verifies the positions-and-rotations API agrees with positioned-text results.
- Checks regions include upright lines and exclude boxes incorrectly shifted left.
- Asserts bounds and rotations for 180-degree, horizontal, and vertical reflections.

Closes #515.

<sup>Written for commit a4893b4e333cfc9e69eee95d176ca79829816b79. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/pdf-inspector/pull/549?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


abimaelmartell:
@cubic-dev-ai review this PR

PR #555, opened 2026-09-20
Make the OCR execution provider selectable (DirectML)

Aut0matas:
## What

Makes the OCR execution provider selectable at runtime, so OCR can run on a local GPU.

## Why

`oar-ocr` — the OCR backend this crate uses — supports `cuda` / `tensorrt` / `directml` / `webgpu` /
`coreml` / `openvino` and exposes `OrtSessionConfig::with_execution_providers`. But `Cargo.toml`
builds it with `["simd"]` only, and `ocr_session_config()` never calls that method, so every ONNX
Runtime session lands on the CPU default and the backend's accelerator support is unreachable from
this crate.

I checked `abi/local-ocr-oar` and the open OCR PRs before opening this — `execution_provider`
appears zero times in `src/vision/oar.rs` on that branch, so I don't believe this is already
covered. Happy to be corrected if I missed something.

## Change

Two files:

| File | Change |
|---|---|
| `Cargo.toml` | `oar-ocr` features `["simd"]` → `["simd", "directml"]` |
| `src/vision/oar.rs` | `ocr_session_config()` reads `PDF_INSPECTOR_OCR_EP`; `directml[:N]` appends `[DirectML{device_id:N}, CPU]` |

CPU stays the default, so a build or a machine without the DirectML runtime behaves exactly as it
does today. Additive, not a behaviour change.

One detail worth flagging in review: **provider order is load-bearing.**
`has_accelerator_provider()` inspects only the first entry, and a CPU provider listed first claims
nearly every node before the accelerator gets a chance — so DirectML has to come first.

## Measurements

PP-OCRv6 Small, `--ocr force`, same page range, minimum of two runs per configuration
(Windows 11, RTX 5090 Laptop):

| pages | CPU | DirectML |
|---|---|---|
| 1 | 2,647 ms | 3,907 ms |
| 2 | 3,467 ms | 4,302 ms |
| 4 | 6,258 ms | 7,801 ms |
| 8 | 12,547 ms | 11,948 ms |

**The honest version: the GPU is not where the time goes.**

- DirectML pays roughly 1.3 s of extra session setup, so it *loses* on short jobs and wins only
  about 5% at 8 pages.
- Raising `--ocr-dpi` from 150 to 300 — 4x the detector's input pixels — moves it from 0.99x to
  1.02x. Even with the inference work multiplied it does not pull ahead, which suggests the
  per-page cost is dominated by PDFium rendering and CPU-side preprocessing rather than by ONNX
  inference.
- Confirmed this is not a silent fallback: the GPU peaked at 28% utilization, and CPU vs DirectML
  output is **byte-identical**.

So I am not claiming GPU OCR is faster — it mostly isn't. I'm submitting this because provider
selection is currently unreachable at all, and the same hook is what a CUDA / TensorRT / WebGPU /
CoreML backend would need.

## Open questions for the maintainers

- Would you rather the `directml` feature be gated behind a Cargo feature rather than always on? I
  enabled it unconditionally because the runtime switch is the point, but it does add dependency
  edges for builds that will never use a GPU. Easy to change if you prefer.
- I only wired DirectML because that is what I can test on Windows. The same pattern extends to the
  other providers, but I can't verify cuda/tensorrt/coreml here, so I left them out.

---
## Summary by cubic
Makes the OCR execution provider selectable at runtime so OCR can run on a local GPU via DirectML. CPU stays the default unless `PDF_INSPECTOR_OCR_EP` is set.

**Review notes**
- `PDF_INSPECTOR_OCR_EP=directml[:N]` selects DirectML on adapter `N` with CPU fallback; `N` defaults to `0`.
- DirectML is gated behind an explicit `ocr-directml` Cargo feature, so `--features ocr` alone still builds and stays on CPU.
- Malformed or negative adapter IDs fall back to CPU instead of silently selecting adapter 0.
- DirectML must be listed before CPU because `has_accelerator_provider()` only inspects the first provider.
- DirectML adds roughly 1.3s of session setup, so it loses on short jobs and only gains about 5% at 8 pages.
- Output is byte-identical between CPU and DirectML, and GPU utilization was confirmed, so this isn't a silent fallback.
- DirectML requires DirectML-enabled `onnxruntime.dll` and `DirectML.dll` on the search path.

<sup>Written for commit ad7c75a210d06fbd52053899565a36bf502134bf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/pdf-inspector/pull/555?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Aut0matas:
Both findings were valid — fixed in ad7c75a.

**1. Malformed adapter id silently coerced to device 0**

Correct, and worth fixing precisely because adapter ordering is not guessable: on this machine
`EnumAdapters1` puts the discrete GPU at 0, but on a hybrid laptop the integrated GPU commonly
enumerates first. So `directml:1x` quietly running on adapter 0 could mean a different device than
intended.

The parse is now a pure function that rejects malformed and negative ids and falls back to the CPU
provider:

```rust
fn parse_directml_device(value: &str) -> Option<i32> {
    let (name, arg) = match value.split_once(':') {
        Some((name, arg)) => (name, Some(arg)),
        None => (value, None),
    };
    if !matches!(name.trim().to_ascii_lowercase().as_str(), "directml" | "dml") {
        return None;
    }
    match arg {
        None => Some(0),
        Some(device) => device.trim().parse::<i32>().ok().filter(|id| *id >= 0),
    }
}
```

Unit-tested for `directml`, `dml`, `DML:2`, whitespace, `directml:1x`, `directml:-1`, `directml:`,
`directml:0x1`, `""`, `cpu`, `cuda:0`.

**2. DirectML could not be turned off, and was forced on every `ocr-oar` consumer**

Also correct. I had hardcoded `["simd", "directml"]` on the `oar-ocr` dependency *and* referenced
`OrtExecutionProvider::DirectML` unconditionally, so disabling the feature broke compilation — the
feature was decorative. Since `ort`'s `directml` feature is not target-gated upstream, that also
pushed `ort/directml` onto CPU-only consumers.

DirectML now lives behind its own feature, deliberately **not** folded into `ocr`:

```toml
ocr-directml = ["ocr-oar", "oar-ocr/directml"]
```

so `--features ocr` alone still builds and stays on the CPU provider, and GPU support is opt-in
with `--features ocr,ocr-directml`. The `OCR_EP_ENV` constant and the `OrtExecutionProvider` import
are `#[cfg]`-gated to match.

**Verification**

- `cargo build --release` passes with both `--features ocr` and `--features ocr,ocr-directml`
- `cargo clippy -- -D warnings` passes with both (zero warnings)
- `cargo fmt --check` clean

One caveat: `cargo test` cannot link on my machine — MinGW GNU ld fails on this crate's cdylib with
`error: export ordinal too large: 172977`, which is unrelated to this change (release builds link
fine). I verified the parser by compiling the same function standalone and running the assertions,
15/15 passing, but the in-repo test itself has only been exercised that way. If CI runs on
Linux/macOS that gap closes immediately.

If you would rather have one commit than a fixup, squash on merge — no objection.

